PR #7, opened 2025-10-06
refactor!

gipert:
@tdixon97 there is a specification of the metadata in the tests directory, let me know if you like it. Also have a look at the code to build the remage command line.


simflow-config.yaml
```diff
@@ -1,22 +1,18 @@
-experiment: l200a
-runlist:
-- l200-p03-r000-phy
```
tdixon97, line 3: i think this often causes bugs in the past, shouldnt this be in legend-datasets?

tests/dummyprod/inputs/simprod/config/tier/stp/l200p03/simconfig.yaml
```diff
@@ -0,0 +1,29 @@
...
+  generator: ~defines:Rn222-to-Po214
+  confinement: ~defines:hpge-bulk
+  primaries_per_job: 100_000
+  number_of_jobs: 30
+
+l200p03-hpge-surface-K42:
+  template: $_/template.mac
+  generator: ~defines:Ra228-to-Ac228
+  confinement:
+    - ~volumes.surface:V.*
```
tdixon97, line 26: i am not sure i get this, doesnt it need revertex (only n+)

tests/dummyprod/simflow-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    stp: 5000
+    ver: 10000
+
+paths:
+  benchmarks: $_/generated/benchmarks
+  log: $_/generated/log
+
+  metadata: $_/inputs
+  config: $_/inputs/simprod/config
+  optical_maps:
```
tdixon97, line 15: are these still considered input not generated?

simflow-config.yaml
gipert: what bugs?

tests/dummyprod/inputs/simprod/config/tier/stp/l200p03/simconfig.yaml
gipert: yeah it's just for testing, i forgot to update names

simflow-config.yaml
tdixon97: its easy to forgot one run, its better to take that info from a central location (if possible), i guess we should support both

tests/dummyprod/simflow-config.yaml
gipert: i don't think it makes sense to produce them in the same workflow, we'll have a separate one

simflow-config.yaml
gipert: in the long term we can move this to the metadata

PR #9, opened 2025-10-07
HPGe drift time maps

gipert:
- [x] add Snakemake rule(s)
- [x] configure Julia environment
- [ ] perform large scale test on NERSC
- [ ] add rule to crawl the slow control?

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,45 @@
+style = "default"
+
+indent = 4
+margin = 120
+
+always_for_in = true
+
+short_to_long_function_def = true
+long_to_short_function_def = false
+force_long_function_def = false
+always_use_return = true
+
+remove_extra_newlines = false
+
+align_assignment = true
+align_struct_field = true
+align_pair_arrow = true
+align_conditional = true
+
+disallow_single_arg_nesting = false
+
+trailing_comma = false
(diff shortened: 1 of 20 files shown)


gipert:
we could store DT maps for different angles in the same file? something to think about

gipert:
@tdixon97 could you review this and think about how to store maps for different angles?

tdixon97:
I do not think the storing of multiple angles is so challlenging, lets just do

```
FILENAME/
├── OBJ1 · struct{r, z, FIELD}
│   ├── r      · array<1>{real}  ── {'units': 'UNITS'}
│   ├── z      · array<1>{real}  ── {'units': 'UNITS'}
│   └── FIELD  · array<2>{real}  ── {'units': 'UNITS'}
└── OBJ2 · struct{r, z, FIELD}
    ├── r      · array<1>{real}  ── {'units': 'UNITS'}
    ├── z      · array<1>{real}  ── {'units': 'UNITS'}
    └── FIELD  · array<2>{real}  ── {'units': 'UNITS'}
    
```
I think the reboost code is pretty flexible to this choice, we can also have multiple "fields" inside the same struct but I do not like it so much

gipert:
yeah i don't have a strong opinion. in principle also multiple fields in the same structure is supported in reboost right?

PR #11, opened 2025-10-26
Document Snakemake rules in API docs

gipert:


gipert:
@ggmarshall the script can be moved to legend-dataflow-scripts and we can have this in legend-dataflow too

ggmarshall:
noice, do you want to do that?

gipert:
mmmmmh no

PR #17, opened 2025-10-29
NERSC-related changes

gipert:


gipert:
@ManuelHu @tdixon97 in https://github.com/legend-exp/legend-simflow/pull/17/commits/d3db3119a670a628b694efee086ef50c517d33d4 I have added a first working version of the build_hit() script that convolves the optical map. You'll see that it will be rather straightforward to add more processors

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -0,0 +1,83 @@
...
+
+import lgdo
+import pyg4ometry
+import pygeomtools
+import reboost.spms
+from lgdo import lh5
+from lgdo.lh5 import LH5Iterator
+
+
+def get_sensvols(geom, det_type: str | None = None) -> list[str]:
```
ManuelHu, line 25: should we add such a filtering capability into pygeomtools?

gipert: yes, but maybe let's first write this script completely in the next weeks. at the end we can think about what is worth moving to pygeomtools

PR #26, opened 2025-11-20
Documentation and update some paths

gipert:


docs/source/manual/meta.md
gipert, line 1: @tdixon97

PR #30, opened 2025-12-02
Run partitioning

gipert:


workflow/src/legendsimflow/scripts/make_simstat_partition_file.py
```diff
@@ -0,0 +1,126 @@
...
+# ...
+#
+# where the number of events in each file is partitioned in runs, such that the
+# global partitioning in n_events_part is respected.
+
+# remaining events per run to allocate across jobs
+remaining_run_events = dict(n_events_part)
+
+# job -> run -> [start, end]
+job_partitions: dict[str, dict[str, list[int]]] = {}
```
tdixon97: I think all the following should be a (tested) function

gipert: i can do that

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -52,180 +56,102 @@
...
+            fccd = mutils.get_sanitized_fccd(metadata, det_name)
+            det_loc = geom.physicalVolumeDict[det_name].position
+            dt_map = reboost_utils.load_hpge_dtmaps(snakemake.config, det_name, runid)  # noqa: F821
+
+            # iterate over input data
+            for lgdo_chunk in iterator:
+                chunk = lgdo_chunk.view_as("ak")
+                _distance_to_surf = AttrsDict()
+
+                for surf in ("nplus", "pplus", "passive"):
```
tdixon97: the latter 2 are maybe not used right now?

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -52,180 +56,102 @@
...
+                        det_loc.eval(),
+                        distances_precompute=chunk.dist_to_surf * 1000,
+                        precompute_cutoff=(fccd + 1),
+                        surface_type="nplus",
+                    )
+
+                _activeness = reboost.math.functions.piecewise_linear_activeness(
+                    _distance_to_surf.nplus,
+                    fccd=fccd,
+                    dlf=0.2,
```
tdixon97, line 128: should be from some pars file

workflow/src/legendsimflow/reboost.py
```diff
@@ -20,13 +20,26 @@
...
 from lgdo import LGDO, lh5
 
+from . import patterns
+from .utils import SimflowConfig
+
 log = logging.getLogger(__name__)
 
 
 def make_output_chunk(chunk: LGDO) -> lgdo.Table:
+    """Prepare output detector table chunk for the hit tier.
```
tdixon97, line 37: We will need to adapt this for new (ak.Array) processor conventions?

workflow/src/legendsimflow/reboost.py
```diff
@@ -50,12 +63,31 @@ def make_output_chunk(chunk: LGDO) -> lgdo.Table:
...
-    wo_mode = "append_column" if chunk_idx == 0 else "append"
+    wo_mode = (
+        "append_column"
+        if objname.strip("/") not in lh5.ls(outfile, "hit/")
+        else "append"
+    )
+
+    # add runid column to chunk
+    # NOTE: is this ok?
+    if runid is not None:
```
tdixon97: this is good but I do not think it should be done in an IO function, I think it should be another processor.

workflow/src/legendsimflow/reboost.py
```diff
@@ -87,3 +119,124 @@ def get_sensvols(geom, det_type: str | None = None) -> list[str]:
...
+    else:
+        msg = (
+            f"no valid time maps found for {det_name} in {hpge_dtmap_file}, "
+            "drift time will be set to NaN"
+        )
+        log.warning(msg)
+        dt_map = None
+
+
+def get_remage_hit_range(
```
tdixon97, line 147: Docstring would be nice here, also to explain a bit how the function works

workflow/src/legendsimflow/scripts/make_simstat_partition_file.py
```diff
@@ -0,0 +1,126 @@
...
+runinfo = snakemake.params.runinfo  # noqa: F821
+output_file = Path(snakemake.output[0])  # noqa: F821
+log_file = snakemake.log[0]  # noqa: F821
+metadata = snakemake.config.metadata  # noqa: F821
+runlist = sorted(snakemake.params.runlist)  # noqa: F821
+
+# setup logging
+log = ldfs.utils.build_log(metadata.simprod.config.logging, log_file)
+
+# get full simulation event statistics
```
tdixon97, line 41: still a bit of (untested) logic here

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -52,180 +56,102 @@
...
-            )
 
-    # process the scintillator output
-    if geom_meta.detector_type == "scintillator" and det_name == "lar":
-        log.info("processing the 'lar' scintillator table...")
+                energy = ak.sum(chunk.edep * _activeness, axis=-1)
 
-        for lgdo_chunk in iterator:
-            chunk = lgdo_chunk.view_as("ak")
+                if dt_map is not None:
```
tdixon97, line 133: Wouldnt it be better to move this logic  into the reboost function? All this if statements are a bit ugly? Or we could refactor into one function like `get_psd_fields` or so?

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -0,0 +1,146 @@
...
+
+# setup logging
+log = ldfs.utils.build_log(metadata.simprod.config.logging, log_file)
+
+# load the geometry and retrieve registered sensitive volumes
+geom = pyg4ometry.gdml.Reader(gdml_file).getRegistry()
+sensvols = pygeomtools.detectors.get_all_sensvols(geom)
+
+# loop over the sensitive volumes registered in the geometry
+for det_name, geom_meta in sensvols.items():
```
tdixon97, line 51: Make a function

workflow/src/legendsimflow/reboost.py
gipert: sure, once there's a new reboost release

workflow/src/legendsimflow/scripts/tier/hit.py
gipert: i think this script is the right place to supply defaults for non-existent fields?

gipert: in the long term, if we add more PSD fields, we can refactor in a separate function. I don't see the necessity for now

workflow/src/legendsimflow/scripts/tier/opt.py
gipert: not sure

workflow/src/legendsimflow/scripts/tier/hit.py
gipert: let's change this to the proper thing once we have a set of new dlf/fccds, now it's a bit silly to make a config for 1 param

workflow/src/legendsimflow/scripts/tier/opt.py
tdixon97: how is it tested otherwise?

gipert: what do you want to test exactly?

tdixon97: Your right its a simple script, but still it has some logic

workflow/src/legendsimflow/scripts/tier/hit.py
tdixon97: Sure, I think it would be great, all the if statements are ugly

PR #35, opened 2025-12-16
NERSC-related changes

gipert:


gipert:
@tdixon97 the last commit is an example of how to add a validation plot

PR #37, opened 2025-12-20
feat: extract A/E model parameters and compute it in hit.py

tdixon97:
This currently only works on the PRL data, supporting also more recent p14-p16 datasets is possible, but supporting both at the same time is ugly (maybe neccesary for now)

- [x] add plot targets to "all" rule?
- [x] docs
- [x] compute A/E in hit.py
- [x] rename valid hpge lookup routines


workflow/src/legendsimflow/hpge_pars.py
gipert, line 1: pars.py is too generic... maybe hpge_pars.py?

tdixon97:
Well my idea was more stuff would go here

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

tdixon97:
@gipert I added some tests, but what I could do is a bit limited with the test-data we have. Eg. it was not really possible to test that both old and new formats will work.
But maybe its enough for now and we can try to integrate this into the simflow?

workflow/src/legendsimflow/aggregate.py
```diff
@@ -139,6 +139,10 @@ def gen_list_of_hpges_valid_for_dtmap(config: SimflowConfig, runid: str) -> list
 
     hpges = []
     for _, hpge in chmap.group("system").geds.items():
+        # TEMPORARY HACK
+        if hpge.name in ("V00050A", "V13046A", "V00048B", "V14654A"):
```
tdixon97, line 149: I had a quick look, the latter 3 detectors were well below depletion voltage in p16. We should probably skip PSD sims for any AC detectors. I will look into 50A

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -128,20 +138,45 @@
...
+                            **currmod_pars,
+                        )
+                        # and calculate A/E
+                        a_max = reboost.hpge.psd.maximum_current(
+                            chunk.edep_active,
+                            drift_time,
+                            template=a_tmpl,
+                            times=np.arange(t_domain["low"], t_domain["high"]),
+                        )
+                        aoe = a_max / energy
```
gipert: @tdixon97 is this ok?

tdixon97:
No it is not quite right.

  *
mean_aoe is not a parameter of the current template function
  *
I typically set A_max to 1, to avoid needing to recalibrate A/E later.

Also wouldn't it be better to have a boolean flag like "simulate_psd" or so instead of guessing from the pars being None (for readability).

I'd also move the simulation of all PSD stuff into a function to clean up the script .







Sent from Outlook for Android<https://aka.ms/AAb9ysg>

gipert:
ok, now the template max should be at 1. i moved some stuff in a separate function, otherwise i think the hit script looks readable at the moment, we can factor out more in the future if it becomes longer.

PR #39, opened 2025-12-22
add plots of stp tier event vertices

gipert:
image

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -74,8 +74,7 @@ def gen_target_all():
         elif "stp" in make_tiers:
             return (
                 rules.gen_all_tier_stp.input,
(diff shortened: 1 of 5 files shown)


gipert:
comments @tdixon97?

tdixon97:
I think its good, are there any other ways to make this more useful? Eg. using the pygeomtools gdml viewer?

gipert:
yes i thought the same, it would be maybe useful to overlay some geometry elements (like the detectors). but i don't know exactly how to do it, would take some time for me

tdixon97:
I think it is a nice enhancement rather than essential. The most likely bugs will be some part missing from the sampling (which your plots will show us). 
The containment check in remage should already catch larger issues.

PR #46, opened 2026-01-08
cache list of valid HPGe detectors

gipert:
This is going to massively speedup the Simflow initialization.


tdixon97:
How much is the speedup?

gipert:
i'm not done yet, i have to implement the caching

gipert:
speedup 690%

PR #55, opened 2026-01-16
evt tier

gipert:
- **improve test_aggregate.py**
- **add infrastructure for evt tier**


gipert:
Should be all tested and working. I would merge this and keep adding functionality in future PRs.

PR #56, opened 2026-01-16
basic evt building (only forward)

tdixon97:
- Does not yet handle the optics
- Statuses should be added to hit
- Might want to write also this name --> uid mapping
- add psd subtable?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ dependencies = [
     "pyg4ometry",
     "pygama >=2.2.3",
     "pylegendmeta >=1.3.5",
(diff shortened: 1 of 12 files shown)


tdixon97:
I found some issues we need to resolve:
- Often the FCCD value is not found in the hit tier (lots of detectors use the defaults)
- The drift times are always NaN --> NaN for both `dt_heuristic` and `aoe`
- https://github.com/legend-exp/reboost/blob/5d57bf6e0abb484986b5210f43c498e6856434da/src/reboost/utils.py#L448 does not work on `/dvs_ro/` on NERSC due to file locking.
-

PR #57, opened 2026-01-19
Add julia script for ideal Pulse Shape Library generation

SalehGiovanna:
Julia script to build ideal Pulse Shape Library using ssd

_Realistic PSL construction from ideal (py)_ ---> Moved to separate PR

### DIFF
--- a/workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl
+++ b/workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl
@@ -21,9 +21,47 @@ using Unitful
 using Base.Threads
 using LinearAlgebra
 using RadiationDetectorDSP
+using LegendDataManagement
+using Printf
+using PropDicts
 
 # 8-connectivity neighbor offsets for map extension
 const NEIGHBOR_OFFSETS_8CONN = [(-1, -1), (-1, 0), (-1, 1), (0, -1), (0, 1), (1, -1), (1, 0), (1, 1)]
+
+
+"""
+    load_detector_metadata(meta_path::String, det::String, opv_val::Union{Float32, Nothing})
+
+Load detector and crystal metadata from legend-metadata, and set the operational voltage.
+
+# Arguments
+- `meta_path`: Path to legend-metadata
+- `det`: HPGe detector name (e.g. "V03422A")
+- `opv_val`: Operational voltage in V (Float32). If `nothing`, the value from metadata is used.
+
+# Returns
+- `Tuple`: `(meta, xtal, opv_val)` where `meta` is the detector metadata (PropDict),
+  `xtal` is the crystal metadata (PropDict), and `opv_val` is the operational voltage
+  as Float32 (either from metadata or as provided)
+"""
+function load_detector_metadata(meta_path::String, det::String, opv_val::Union{Real,Nothing})
+
+    meta = readprops("$meta_path/hardware/detectors/germanium/diodes/$det.yaml")
+
+    if isnothing(opv_val)
+        opv_val = Float32(meta.characterization.l200_site.recommended_voltage_in_V)
+        @info "No OPV provided. Using metadata default: $opv_val V"
+    else
+        @info "Using user-provided OPV: $opv_val V"
+    end
+    meta.characterization.l200_site.recommended_voltage_in_V = opv_val
+
+    ids = Dict("bege" => "B", "coax" => "C", "ppc" => "P", "icpc" => "V")
+    crystal = ids[meta.type] * @sprintf("%02d", meta.production.order) * meta.production.crystal
+    xtal = readprops("$meta_path/hardware/detectors/germanium/crystals/$crystal.yaml")
+
+    return meta, xtal, opv_val
+end
 
 
 """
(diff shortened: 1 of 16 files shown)


output
```diff
@@ -0,0 +1 @@
+/pscratch/sd/g/giosal/legend-simflow/output
```
tdixon97: whats this?

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,325 @@
+# Script to build ideal waveform map of a detector using ssd
+# Map construction based on https://github.com/legend-exp/legend-simflow/blob/main/workflow/src/legendsimflow/scripts/make_hpge_drift_time_maps.jl
```
tdixon97: I wonder how much can be shared? Some of the functions are the same

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,325 @@
...
+        "--metadata"
+            help = "Path to legend-metadata"
+            required = true
+        "--output-file"
+            help = "Path to output LH5 file"
+            required = true
+        "--opv"
+            help = "detector operational voltage in V (defaults to metadata value)"
+            #required = true # Make sure thie script is using this!!
+        "--use-sqrt" # Outdated??
```
tdixon97: I would remove

output
SalehGiovanna: Ah no that shouldn't have been part of the commit let me remove those

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
SalehGiovanna: Yes we should make a library with the functions and then the scripts pick from there

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
SalehGiovanna: I wanted to make sure with @hervasa2 that this is really no more the way to include the temperature model

tdixon97: I wrote this not David, for now we are still using the old method, see how its done in the drift time map code.

tdixon97:
Can you run pre-commit locally, this will include checks on the julia code and formatting

workflow/src/legendsimflow/scripts/make_hpge_realistic_wf_map.py
```diff
@@ -65,9 +68,7 @@ def create_response_kernel(dt, mu_bandwidth, sigma_bandwidth, tau_rc, gaussian_o
     rf_preamp /= np.sum(rf_preamp)
 
     # Convolve preamp and digitizer responses - 'full' mode results in a length of (len(rf_preamp) + len(rf_digi) - 1)
-    rf = convolve(rf_preamp, rf_digi, mode='full')
-    
-    return rf
+    return convolve(rf_preamp, rf_digi, mode='full')
 
 
 def convolve_waveforms(wf_array, rf_kernel):
```
tdixon97: would be great to add here type hints and docstrings, and move them into a module (ie not in `scripts`) and add tests.

workflow/src/legendsimflow/scripts/make_hpge_realistic_wf_map.py
```diff
@@ -0,0 +1,470 @@
...
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
+
+ALIGNMENT_IDX = 2000  # Index to align current waveforms to Amax ######################### Want to put 1000 as default???
+NSAMPLES_OUTPUT_CURRENT_WFS = (
+    4001  # Final length of the realistic current waveforms in the map
+)
+
+
+def create_response_kernel(
```
tdixon97: add type hints here

tdixon97: I think almost all these functions do not belong in the `scripts` subdirectory

workflow/src/legendsimflow/scripts/make_hpge_realistic_wf_map.py
```diff
@@ -0,0 +1,470 @@
...
+    dt, mu_bandwidth, sigma_bandwidth, tau_rc, gaussian_only=False
+):
+    """
+    Create the system response kernel (gaussian + exponential decay)
+
+    Parameters
+    ----------
+    dt : float
+        The time step between samples in the waveform (in ns)
+    mu_bandwidth : float
```
tdixon97: If you add the type hints, the type annotations in the docstring can be removed (theyll be added automatically)

workflow/src/legendsimflow/scripts/make_hpge_realistic_wf_map.py
```diff
@@ -0,0 +1,470 @@
...
+        Array of filtered waveforms - same shape as input!
+    """
+
+    mwa_currents = np.zeros_like(
+        wf_array, dtype=float
+    )  # Risk to saturate memory?? Should check this??
+
+    ########## Should vectorize?? ##########
+    # If run on map only, this is fast enough
+    for i in range(len(wf_array)):
```
tdixon97: You do not need the loop you can act directly on a 2D array of wfs

workflow/src/legendsimflow/scripts/make_hpge_realistic_wf_map.py
```diff
@@ -0,0 +1,470 @@
...
+            start_dst = start_src + shift
+            end_dst = end_src + shift
+
+            shifted_wfs[i, start_dst:end_dst] = wfs[i, start_src:end_src]
+
+    return shifted_wfs, peak_indices
+
+
+def generate_realistic_map(
+    ideal_wf_map_obj, rf_kernel, Amax_alignment_idx=ALIGNMENT_IDX
```
tdixon97: I think having the default be a global variable is not considered good practice

SalehGiovanna: I think now this is fine? It's global just in the script - not in the module

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
gipert: i have to change some things in the dt map script soon, maybe i can take the opportunity to refactor these julia functions to a small library and then you can base your work here on it

gipert:
see https://github.com/legend-exp/legend-simflow/pull/91

SalehGiovanna:
@gipert I saw it, but I am not using those functions in my script, even though its true that there is a lot of shared code.
So the possibilities I see are:

- Merge this and #91 separately and then modify the functions in the libjl to be usable both for the dt map and for the wfs map [best solution in my opinion]
- Add my functions for the wfs map construction (currently in make_hpge_ideal_wf_map.jl) to libjl and modify my script to include them from there

gipert:
Yes, the first is the right solution

tdixon97:
@SalehGiovanna I rebased this onto #95. I think its in a great shape, I can quickly make the last fixes then we are done with this (we can add tests in a further PR)

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+using LegendDataManagement
+using Unitful
+using LegendHDF5IO
+using Base.Threads
+using ArgParse
+using PropDicts
+using Printf
+using RadiationDetectorDSP
+
+include(joinpath(@__DIR__, "libjl", "drift_time_helpers.jl"))
```
tdixon97:
```suggestion
using LegendSimflow
```

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+        action = :store_true
+        "--use-sqrt-new" # REVIEW: Outdated??
+        help = "Use square-root temperature model with 2016 inputs"
+        action = :store_true
+        "--only-holes"
+        help = "Use only the hole contribution"
+        action = :store_true
+        "--use-bulk-drift-time"
+        help = "Use the drift time for the nearest bulk point for surface events"
+        action = :store_true
```
tdixon97:
```suggestion
      
```
We should remove all these options (see the dt map script)

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+    end
+
+
+
+
+    if (!handle_nplus)
+        meta.production.characterization.manufacturer.dl_thickness_in_mm = 0.1
+        meta.characterization.combined_0vbb_analysis.fccd_in_mm.value = 0.1
+    end
+
```
tdixon97:
```suggestion
```
I would remove this (old) option

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+    crystal = ids[meta.type] * @sprintf("%02d", meta.production.order) * meta.production.crystal
+    xtal = readprops("$meta_path/hardware/detectors/germanium/crystals/$crystal.yaml")
+
+    if !use_corrections
+        if :corrections in keys(xtal.impurity_curve)
+            xtal.impurity_curve.corrections.scale = 1
+        else
+            @warn "Impurity corrections were requested to be disabled, but no corrections are defined for crystal $crystal."
+        end
+    end
```
tdixon97:
```suggestion
```
I would also remove this (see dt map script)

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+    if (use_sqrt)
+        charge_drift_model = ADLChargeDriftModel(
+            joinpath(
+                SolidStateDetectors.get_path_to_example_config_files(),
+                "ADLChargeDriftModel",
+                "drift_velocity_config_squareroot.yaml"
+            )
+        )
+        sim.detector = SolidStateDetector(sim.detector, charge_drift_model)
+    end
```
tdixon97:
```suggestion
    charge_drift_model = ADLChargeDriftModel(
        "$meta_path/simprod/config/pars/geds/ssd/adl-2016-temp-model.yaml"
    )
    sim.detector = SolidStateDetector(sim.detector, charge_drift_model)

    
```

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+        sim,
+        sim.detector.contacts[1].id,
+        refinement_limits = [0.2, 0.1, 0.05, 0.01],
+        verbose = false
+    )
+
+    # Compute and save ideal waveforms map for different angles
+    output = Dict{Symbol,Any}()
+    for a in CRYSTAL_AXIS_ANGLES
+        eff_angle = use_sqrt ? (a - 45) : a
```
tdixon97:
```suggestion

```
remove?

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+    end
+    meta.characterization.l200_site.recommended_voltage_in_V = opv_val
+
+    handle_nplus = parsed_args["use-bulk-drift-time"]
+    use_sqrt = parsed_args["use-sqrt"]
+    only_holes = parsed_args["only-holes"]
+    use_corrections = parsed_args["use-corrections"]
+    if !use_corrections
+        @warn "Impurity corrections disabled! This may produce unrealistic results."
+    end
```
tdixon97:
```suggestion

```

tdixon97 (commented):
Just some small comments to remove the outdated options, this would also align with the dt map script

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,372 @@
...
+## Constants
+GRID_SIZE = 0.0005 # cylindrical (r, z) grid spacing in meters
+CRYSTAL_AXIS_ANGLES = [0, 45] # crystal axis orientation angles in degrees (<001>, <110>)
+SIM_ENERGY = 2039u"keV" # mono-energetic deposition used to generate ideal waveforms
+MAX_NSTEPS = 3000 # maximum number of integration steps per drift trajectory (dimensionless)
+WAVEFORM_LENGTH = 3000 # number of time samples per simulated waveform (dimensionless)
+TIME_STEP = 1u"ns" # time step between waveform samples
+
+
+## Waveform map construction
```
tdixon97: Can you move this to the drift_time_helpers?

gipert:
after #95 got merge the situation got messy, i work on a rebase now

tdixon97:
@gipert now the commits are just by you and Claude, do you know how to fix this to be Giovanna?
I suppose its due to rebasing?
By the way @SalehGiovanna and I discussed a refactoring to resolve almost all comments, she will implement it today I think

SalehGiovanna:
@tdixon97 yes i'm doing it now, I did not commit anything yet because I saw there was ongoing work, now i do it

gipert:
I fixed the authors. I just have a chat with @SalehGiovanna, she is gonna pick the work up from the current status here.

tdixon97:
Now the authors is me and you Luigi? Should be Giovanna

gipert:
The first commit is giovanna's work and she's an author there

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
```diff
@@ -248,6 +248,132 @@ function find_valid_spawn_position(candidate_idx, spawn_positions, detector; ver
...
+)
+    @info "Computing waveform map at angle $angle_deg deg..."
+
+    SSD = SolidStateDetectors
+    angle_rad = deg2rad(angle_deg)
+
+    # Simulation parameters
+    sim_energy = 2039u"keV"
+    max_nsteps = 3000
+    waveform_length = 3000
```
hervasa2: This is a bit low, in particular when we have large IC's operating close to depletion. I would set both to 5000.

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
```diff
@@ -248,6 +248,132 @@ function find_valid_spawn_position(candidate_idx, spawn_positions, detector; ver
...
+        e = SSD.Event([p], [sim_energy])
+        simulate!(e, sim, Δt = time_step, max_nsteps = max_nsteps, verbose = false)
+
+        if only_holes
+            wf = get_electron_and_hole_contribution(e, sim, 1).hole_contribution
+        else
+            wf = e.waveforms[1]
+        end
+
+        wf_signals_threaded[i] = ustrip(wf.signal)
```
hervasa2: replace with 
```julia
wf_signals_threaded[i] = ustrip(add_baseline_and_extend_tail(wfs_raw, 0, waveform_length).signal)
```

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
```diff
@@ -248,6 +248,132 @@ function find_valid_spawn_position(candidate_idx, spawn_positions, detector; ver
...
+        # Normalize charge waveforms so that their amplitude (energy) = 1
+        max_val = maximum(raw_signal)
+        signal = max_val > 0 ? raw_signal ./ max_val : raw_signal
+
+        # Clip to fixed length and pad tail with last value
+        len = min(length(signal), waveform_length)
+        wf_padded[1:len, idx[2], idx[1]] = signal[1:len]
+        if len < waveform_length && len > 0
+            wf_padded[(len + 1):end, idx[2], idx[1]] .= signal[len]
+        end
```
hervasa2: Remove. Already performed with suggested change to R345

workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl
```diff
@@ -248,6 +248,132 @@ function find_valid_spawn_position(candidate_idx, spawn_positions, detector; ver
...
+            @warn "find_valid_spawn_position did not return a valid spawn position for index $(in_idx[i]); skipping."
+            wf_signals_threaded[i] = Float32[]
+            continue
+        end
+
+        e = SSD.Event([p], [sim_energy])
+        simulate!(e, sim, Δt = time_step, max_nsteps = max_nsteps, verbose = false)
+
+        if only_holes
+            wf = get_electron_and_hole_contribution(e, sim, 1).hole_contribution
```
hervasa2, line 451: In general the `only_holes` scenario can be optimized by drifting only holes instead of drifting all charges (`simulate!`) and then extracting the hole contribution. This would require a bit of work... I could do it in a future PR, let me know if its worth it.

hervasa2 (changes requested):
Very nice! Please see minor requested changes and comments from the SSD side

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
gipert: I agree, I also was also wondering this

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
hervasa2: Sorry ` wf_padded[1:len, idx[2], idx[1]] = signal[1:len]` should stay but can likely be changed to 
`wf_padded[:, idx[2], idx[1]] = signal`

workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl
gipert: I don't think it's so urgent at the moment. As we now only compute maps for every detector only at the actual voltage values used in LEGEND this step does not take a lot of time in the Simflow

workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl
```diff
@@ -248,6 +248,132 @@ function find_valid_spawn_position(candidate_idx, spawn_positions, detector; ver
...
+- `angle_deg`: Azimuthal angle in degrees for the r-z plane rotation
+- `only_holes`: If true, extract only hole contribution; if false, use full waveform
+- `handle_nplus`: If true, handle positions at n+ contact by finding nearest valid point
+- `grid_size`: Grid spacing in meters
+
+# Returns
+- `NamedTuple`: Contains `:r`, `:z`, `:dt` axes and `:waveform_XXX_deg` 3D array
+  `[time_samples, n_z, n_r]` of normalized waveforms (dimensionless)
+"""
+function compute_waveform_map_for_angle(
```
gipert: I would use `pulse_shape_library` instead of `waveform_map` everywhere?

workflow/src/legendsimflow/scripts/make_hpge_ideal_wf_map.jl
```diff
@@ -0,0 +1,163 @@
...
+
+    parsed_args = parse_args(s)
+
+    det = parsed_args["detector"]
+    meta_path = parsed_args["metadata"]
+    output_file = parsed_args["output-file"]
+
+    isfile(output_file) && error("Output file already exists")
+
+    meta = readprops("$meta_path/hardware/detectors/germanium/diodes/$det.yaml")
```
gipert: it would be good to avoid duplicating code that is already in the drift time map script (like setting up the simulation) by factoring it out in an independent function

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
tdixon97: Really, I would be pretty surprised to see > 3000 ns drift times? Maybe still better to be safe

gipert: you are right, it's maybe just a precaution

hervasa2:
I would move to merge this asap and then to switch to the new ssd on everything on a subsequent PR

tdixon97:
We first need to remove the duplicate code and add (atleast basic) tests. For example even a test the code runs is valuable

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
SalehGiovanna: i originally put 5000, there were reasons (A/E heuristic code in reboost if i remember correclty) why I later changed it to 3000.
@tdixon97 if now 5000 in total is fine for reboost i go back to that

gipert:
important missing thing: extend the PSL outside its bounds as done at the end of the function that makes the dt map

SalehGiovanna:
But then how do we handle those waveforms?

tdixon97:
I added a few (basic) tests, that the ssd code runs

hervasa2:
> But then how do we handle those waveforms?

You don's simulate those waveforms. You just copy the waveform from the nearest pixel inside the detector, just like the dt map (but with a waveform instead of the dt)

SalehGiovanna:
But isn't it better to have waveforms outside the detector volume being jus NaNs, so they are easy to identify later?

gipert:
>But isn't it better to have waveforms outside the detector volume being jus NaNs, so they are easy to identify later?

no, the pixelation will always leave some holes in valid areas of the detector, so we want to extend the domain to be safe

PR #59, opened 2026-01-21
evt-tier dev

gipert:


workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -184,13 +188,21 @@
...
+
             out_table.add_field("energy", lgdo.Array(energy, attrs={"units": "keV"}))
             out_table.add_field("drift_time_heuristic", lgdo.Array(dt_heuristic))
             out_table.add_field("aoe", lgdo.Array(aoe))
 
-            # add strings
-            utils.add_field_string("runid", out_table, runid)
-            utils.add_field_string("usability", out_table, usability)
+            _, period, run, _ = mutils.parse_runid(runid)
+            out_table.add_field(
```
tdixon97: i suppose it could be shorter?

gipert:
we can implement splitting the unified TCM and lar stuff in a future PR

workflow/src/legendsimflow/reboost.py
```diff
@@ -93,21 +93,23 @@ def write_chunk(chunk: lgdo.Table, objname: str, outfile: str, objuid: int) -> N
...
-            fields = [*lgdo.lh5.datatype.get_struct_fields(dt), f"det{objuid}"]
+            fields = [*lgdo.lh5.datatype.get_struct_fields(dt), link_name]
             f["hit/__by_uid__"].attrs["datatype"] = (
                 "struct{" + ",".join(sorted(fields)) + "}"
             )
 
 
-def get_sensvols(geom, det_type: str | None = None) -> list[str]:
-    sensvols = pygeomtools.detectors.get_all_sensvols(geom)
+def get_senstables(geom, det_type: str | None = None) -> list[str]:
```
tdixon97: type for geom?

workflow/src/legendsimflow/reboost.py
```diff
@@ -164,8 +166,8 @@ def get_remage_hit_range(
     uid
         remage unique identifier for detector `det_name`.
     evt_idx_range
-        first and last index of events of interest present in the remage output
-        file.
+        `[first, last)` index of events of interest present in the remage
```
tdixon97: I would be more explicit what you mean by [first, last)

workflow/src/legendsimflow/reboost.py
```diff
@@ -265,22 +267,27 @@ def hpge_max_current_cal(
...
     # use tables keyed by UID in the __by_uid__ group.  in this way, the
     # TCM will index tables by UID.  the coincidence criterium is based
     # on Geant4 event identifier and time of the hits
     # NOTE: uses the same time window as in build_hit() reshaping
     pygama.evt.build_tcm(
-        [(hit_file, r"hit/__by_uid__/*")],  # input_tables
+        [(f, r"hit/__by_uid__/*") for f in hit_files],  # input_tables
         ["evtid", "t0"],  # coin_cols
         hash_func=r"(?<=hit/__by_uid__/det)\d+",
         coin_windows=[0, 10_000],
```
tdixon97: is it always in ns ?

workflow/src/legendsimflow/scripts/tier/evt.py
```diff
@@ -14,65 +14,102 @@
...
+
+FORWARD_FIELD_LIST = ["aoe", "drift_time_heuristic", "t0"]
+ENERGY_THR_KEV = 25
+OFF = encode_usability("off")
+ON = encode_usability("on")
+BUFFER_LEN = "500*MB"
 
 args = nersc.dvs_ro_snakemake(snakemake)  # noqa: F821
 
+stp_file = args.input.stp_file
```
tdixon97: not so nice to have this dependency? Make the DAG ugly?

workflow/src/legendsimflow/scripts/tier/evt.py
```diff
@@ -14,65 +14,102 @@
...
 
 # setup logging
 log = ldfs.utils.build_log(metadata.simprod.config.logging, log_file)
 
-log.info("... extracting remage uid map")
+log.info("building hit+opt unified TCM")
+reboost_utils.build_tcm([hit_file, opt_file], evt_file)
+
+# test that the evt tcm has the same amount of rows as the stp tcm
+assert lh5.read_n_rows("tcm", stp_file) == lh5.read_n_rows("tcm", evt_file)
```
tdixon97, line 54: in general I think it would be nice to replace the assertions with raising the proper error and giving a meaningful error message?

workflow/src/legendsimflow/scripts/tier/evt.py
```diff
@@ -14,65 +14,102 @@
...
 
 # iterate
-it = lh5.LH5Iterator(hit_file, "tcm", buffer_len=buffer_len)
+it = lh5.LH5Iterator(hit_file, "tcm", buffer_len=BUFFER_LEN)
 
+log.info("begin iterating over TCM")
 for tcm_chunk in it:
-    # convert tcm to ak
-    tcm_tmp = tcm_chunk.view_as("ak")
+    tcm_ak = tcm_chunk.view_as("ak")
```
tdixon97, line 75: do we want unit handling? I think its not so critical by this point

workflow/src/legendsimflow/scripts/tier/evt.py
```diff
@@ -14,65 +14,102 @@
...
+    # HPGe table
+    out_table.add_field("geds", Table(size=len(tcm_ak)))
+
+    # first read usability and energy
+    usability = _read_hits(tcm_ak, "usability")
+    energy = _read_hits(tcm_ak, "energy")
+
+    # we want to only store hits from events in ON and AC detectors and above
+    # our energy threshold
+    is_good_hit = (usability != OFF) & (energy > ENERGY_THR_KEV)
```
tdixon97: hmm not sure this is correct... this would probably mean we remove an event with (10 keV and 1 MeV) but we should keep it as a (valid) M1 event

tdixon97 (approved):

workflow/src/legendsimflow/reboost.py
gipert: yes because this comes from remage... let me add some better docs

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: yeah you are right, maybe we don't specify this dependency

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: so my rationale for using assertions instead of error messages is:

- the error is never actually expected (pedantic check) and is something that the user is not likely to cause
- it is something that usually shows up only during development
- it is therefore more directed to developers that actually know what it means
- more concise check than the full if clause + error printing

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: you are right, let me improve this

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: no this would not remove an event? this removes only hits?

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: https://github.com/legend-exp/reboost/issues/150

workflow/src/legendsimflow/scripts/tier/evt.py
tdixon97: no because the QC is `all(is_good_hit)`

gipert: as you can see i'm not saving the evt-level flag at the moment. not even sure it's so necessary.

in any case, the right flag is `all(usability)`

gipert: no that's not true

tdixon97: I had it, I think its useful, I would not remove

tdixon97: oh btw this "is_good_hit" definition is wrong

PR #70, opened 2026-01-29
Forced trigger

### LINKED ISSUE #50: Random coincidences correction to photoelectron times
In the hit tier, we should smear the pe times with gaussian noise (taken from the SiPM calibration pars) and add the random coincidences

tdixon97:
This gives the basic function to create the forced trigger library in a format that https://reboost.readthedocs.io/en/stable/api/reboost.spms.html#reboost.spms.pe.corrected_photoelectrons can use. 

For now we have not performed a forced trigger correction that returns the photon times, rather just the summed pe per channel. I think this is ok to start with.

If we want to instead obtain the nested vector of pe times we need new functionality in reboost.

### DIFF
--- a/workflow/src/legendsimflow/reboost.py
+++ b/workflow/src/legendsimflow/reboost.py
@@ -341,6 +341,68 @@ def hpge_max_current(
     ).view_as("ak")
 
 
+def get_forced_trigger_library(evt_file_path: str | Path) -> ak.Array:
+    """Extract a library of forced trigger events to be used
+    in correcting the SiPM pe spectra.
+
+    This reformats the data to only have forced trigger, not pulser events
+    and then stores only the summed pe (used for corrections).
+
+    Parameters
+    ----------
+    evt_file_path
+        Path to the event tier data.
+
+    Returns
+    -------
+    Array with two fields "summed_pe", the number of
+    pe in each SiPM and "rawids" the SiPM channel numbers.
+    """
+
+    summed_pe = None
+    rawids = None
+
+    files = Path(evt_file_path).glob("*")
+
+    for file in files:
+        is_forced = lh5.read("evt/trigger/is_forced", file).view_as("ak")
+        is_pulser = lh5.read("evt/coincident/puls", file).view_as(  # codespell:ignore
+            "ak"
+        )
+
+        idx = ak.where(is_forced & (~is_pulser))[0].to_list()
+
+        spms = lh5.read(
+            "evt/",
+            file,
+            idx=idx,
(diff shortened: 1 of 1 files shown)


workflow/src/legendsimflow/reboost.py
gipert, line 1: again you have some stuff that was already committed... i suspect you are doing something not correct with git

gipert:
I can have a look at how much work is to do the correction on the times today.

tdixon97:
I do not think we should do it now, it isnt used for the main observables

gipert:
i think we should for each simulated event look for a sample of random coincident pulse times in the data (in a certain window, e.g. 6 us) and just save this as an additional field in the opt tier. i think this would be useful to develop time/pattern based veto classifiers

tdixon97:
Definitely useful but not a main priority, and I think rather specialised just for that application.

 We should start with just summed pe and multiplicity, our main observable defining the lar veto cut.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

gipert:
Well @rosannadeckert is already trying to use simulations to study the classifier. In any case my point is that you already have the machinery to read in the evt data, so it should be quite simple to just make store the pulse information, instead of computing a derived quantity.

tdixon97:
I think we should first do the simplest case to get our background model PDFs, then if @rosannadeckert needs the times, maybe she can contribute, since changes are then also needed in reboost

rosannadeckert:
sure, i can help! what kind of change is needed in reboost?

gipert:
let me have a look at what tobi wrote here so i can advise

tdixon97:
I think the change is rather minimal just append lists not sum

gipert:
@rosannadeckert do you want to work on this?

rosannadeckert:
yes

gipert:
Superseded by #93.

tdixon97:
We should not just start from scratch.. and correction should be in reboost

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

gipert:
#93 includes the code from this PR. we can develop the routines here for convenience and move them to reboost later, as we still have to do for all the reboost.py module

tdixon97:
Ok but then it would have been better to start from my branch

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

gipert:
@rosannadeckert can't push to your branch

tdixon97:
I do not really understand why the function adding the corrections is needed here. reboost can already do it.
If you want to change small things its better making a PR there

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

PR #75, opened 2026-01-30
hit: fix PSD observables

gipert:


gipert:
More than 1% of the drift times is NaN. Maybe it's an out-of-bounds problem but we should investigate, it sounds too much to me

PR #93, opened 2026-02-06
Forced trigger library and correction

rosannadeckert:
We generate forced-trigger–like events from the full dataset, since the actual forced triggers are recorded at a low rate. This allows us to maximize the number of random-coincidence events available for the correction.

The implementation is parameterized by:
- a chosen window length,
- a minimum separation between consecutive windows,
- and the allowed window range within the waveform.

For LEGEND-200 data, the window range is set to [1μs,44μs]. This choice is motivated by the distribution of the average number of photoelectrons (summed over channels) in 16 ns bins across the full waveform (see plot).

Each window is assigned a reference time defined as the window start time. This reference time is subtracted from the event t0 values so that they lie within the interval [0, window length].
Alternatively, one could define the reference time as “window start + 1 µs”, which would shift the t0 values to the interval [−1μs, window length−1μs].

A remaining point to decide is how to handle cases where insufficient random-coincidence data are available for the correction. The current implementation raises an error in this situation. As an alternative, we could allow the same random-coincidence event to be reused multiple times.

image

### DIFF
--- a/workflow/src/legendsimflow/scripts/plots/tier_opt_observables.py
+++ b/workflow/src/legendsimflow/scripts/plots/tier_opt_observables.py
@@ -49,36 +49,63 @@ def fig(table):
     gs_top = outer[0].subgridspec(1, 2, width_ratios=[1, 1])
     gs_bot = outer[1].subgridspec(1, 2, width_ratios=[1, 1])
 
+    usability_vals, counts = np.unique(data.usability, return_counts=True)
+    vals = [mutils.decode_usability(v) for v in usability_vals]
+
     # time
     ax = fig.add_subplot(gs_top[0, 0])
-    # bin width = 16 ns
-    h_time = hist.new.Reg(313, 0, 5008, name="photoelectron $t - t_0$ (ns)").Double()
+    h_time = hist.new.Reg(
+        375, -1000, 5000, name="photoelectron $t - t_0$ (ns)"
+    ).Double()
     dt = data.time - data.t0
     h_time.fill_flattened(dt)
-    plot.plot_hist(h_time, ax, n_nans=n_nans(dt))
-    ax.set_ylabel("counts / 10 ns")
+    plot.plot_hist(h_time, ax, n_nans=n_nans(dt), label="simulated")
+
+    if vals != ["off"] and "rc_time" in data.fields:
+        h_time_rc = hist.new.Reg(
+            375, -1000, 5000, name="photoelectron $t - t_0$ (ns)"
+        ).Double()
+        h_time_rc.fill_flattened(data.rc_time)
+        plot.plot_hist(
+            h_time_rc, ax, n_nans=n_nans(data.rc_time), label="random coincidences"
+        )
+
+    ax.set_ylabel("counts / 16 ns")
     ax.set_yscale("log")
+    ax.legend()
 
     # usability
     ax = fig.add_subplot(gs_top[0, 1])
-    vals, counts = np.unique(data.usability, return_counts=True)
-    labels = [mutils.decode_usability(v) for v in vals]
+
+    raw_vals, counts = np.unique(data.usability, return_counts=True)
+    labels = [mutils.decode_usability(v) for v in raw_vals]
     plt.pie(
         counts,
         labels=labels,
         colors=[plot.USABILITY_COLOR[lab] for lab in labels],
         autopct="%1.1f%%",
     )
+
     ax.set_aspect("equal")  # keep it circular
 
     ax = fig.add_subplot(gs_bot[0, 0])
     energy = data.energy[~data.is_saturated]
 
-    h_peamp = hist.new.Reg(
+    h_peamp_ns = hist.new.Reg(
         300, 0, 20, name="light per cluster (photoelectrons)"
     ).Double()
-    h_peamp.fill_flattened(energy)
-    plot.plot_hist(h_peamp, ax, n_nans=n_nans(energy), label="non-saturated")
+    h_peamp_ns.fill_flattened(energy)
+    plot.plot_hist(h_peamp_ns, ax, n_nans=n_nans(energy), label="non-saturated (sim)")
+
+    if vals != ["off"] and "rc_energy" in data.fields:
+        h_peamp_rc = hist.new.Reg(
+            300, 0, 20, name="light per cluster (photoelectrons)"
+        ).Double()
+        h_peamp_rc.fill_flattened(data.rc_energy)
+        plot.plot_hist(
+            h_peamp_rc, ax, n_nans=n_nans(data.rc_energy), label="random coincidences"
+        )
+
     ax.set_ylabel("counts")
     ax.set_yscale("log")
     ax.legend()
@@ -86,16 +113,27 @@ def fig(table):
     ax = fig.add_subplot(gs_bot[0, 1])
     h_npe = hist.new.Reg(200, 0, 150, name="light per event (photoelectrons)").Double()
     h_npe.fill(ak.sum(energy, axis=-1))
-    plot.plot_hist(h_npe, ax, flow="hint", label="non-saturated")
+    plot.plot_hist(h_npe, ax, flow="hint", label="non-saturated (sim)")
 
     h_sat = h_npe.copy(deep=True)
     h_sat.reset()
     h_sat[hist.overflow] = ak.sum(data.is_saturated)
-    h_sat.plot(ax=ax, label="saturated", flow="show", yerr=False)
+    if h_sat.sum() != 0:
+        h_sat.plot(ax=ax, label="saturated (sim)", flow="show", yerr=False)
 
     ax.set_ylabel("counts")
     ax.legend()
+
+    if vals != ["off"] and "rc_energy" in data.fields:
+        h_npe_rc = hist.new.Reg(
+            200, 0, 150, name="light per event (photoelectrons)"
+        ).Double()
+        h_npe_rc.fill(ak.sum(data.rc_energy, axis=-1))
+        plot.plot_hist(h_npe_rc, ax, flow="hint", label="random coincidences")
+
+    ax.set_ylabel("counts")
     ax.set_yscale("log")
+    ax.legend()
 
     fig.suptitle(f"{simid}: {table} hits")
     return fig
--- a/workflow/src/legendsimflow/scripts/tier/opt.py
+++ b/workflow/src/legendsimflow/scripts/tier/opt.py
@@ -35,7 +35,7 @@ from lgdo.lh5 import LH5Iterator
 from reboost.optmap.convolve import OptmapForConvolve
 
 from legendsimflow import metadata as mutils
-from legendsimflow import nersc
+from legendsimflow import nersc, spms_pars, utils
 from legendsimflow import reboost as reboost_utils
 from legendsimflow.profile import make_profiler
 from legendsimflow.tcm import build_tcm
@@ -50,9 +50,11 @@ gdml_file = args.input.geom
 log_file = args.log[0]
 metadata = args.config.metadata
 optmap_per_sipm = args.params.optmap_per_sipm
+add_random_coincidences = args.params.add_random_coincidences
 scintillator_volume_name = args.params.scintillator_volume_name
 simstat_part_file = args.input.simstat_part_file
 usabilities = AttrsDict(load_dict(args.input.detector_usabilities[0]))
+l200data = args.config.paths.l200data
 
 hit_file, move2cfs = nersc.make_on_scratch(args.config, hit_file)
 
@@ -73,6 +75,11 @@ geom = pyg4ometry.gdml.Reader(gdml_file).getRegistry()
 sens_tables = pygeomtools.detectors.get_all_senstables(geom)
 
 
+def _ak_array_of_empty_arrays(n):
+    content = ak.Array(np.empty(0, dtype=np.float64))
+    return ak.unflatten(content, np.zeros(n, dtype=np.int64))
+
+
 def process_sipm(
     iterator: LH5Iterator,
     optmap_lar: str | Path | OptmapForConvolve,
@@ -81,6 +88,8 @@ def process_sipm(
     out_file: str | Path,
     runid: str,
     usability: str,
+    rc_library: ak.Array | None,
+    rc_offset: dict,
 ) -> None:
     with perf_block("load_optmap()"):
         if not isinstance(optmap_lar, OptmapForConvolve):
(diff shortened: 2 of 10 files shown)


gipert:
If you tell me your favourite evt tier file (with enough data for tests to be meaningful), I can upload a truncated version to legend-testdata so you can add some unit tests

workflow/src/legendsimflow/reboost.py
```diff
@@ -558,3 +558,138 @@ def cluster_photoelectrons(
...
+
+    # determine window starts and ends
+    starts = []
+    current_start = win_range_ns[0]
+    while current_start + win_len_ns <= win_range_ns[1]:
+        starts.append(current_start)
+        current_start += win_len_ns + min_sep_ns  # move to next window
+    ends = [s + win_len_ns for s in starts]
+
+    files = Path(evt_file_path).glob("*")
```
gipert: let's pass a list of files as argument and not a path to a directory

workflow/src/legendsimflow/reboost.py
```diff
@@ -558,3 +558,138 @@ def cluster_photoelectrons(
         cur_a = ak.unflatten(cur_a, parent_counts)
 
     return cur_t, cur_a
+
+
+def get_forced_trigger_library(
+    evt_file_path: str | Path,
+    win_len_ns: float = 6_000,
+    min_sep_ns: float = 6_000,
```
gipert: is this really needed? won't we always want to extract adjacent windows (i.e. win_len_ns == min_sep_ns)?

workflow/src/legendsimflow/reboost.py
```diff
@@ -558,3 +558,138 @@ def cluster_photoelectrons(
...
+    ends = [s + win_len_ns for s in starts]
+
+    files = Path(evt_file_path).glob("*")
+
+    for file in files:
+        spms = lh5.read(
+            "evt/",
+            file,
+            field_mask=["spms/energy", "spms/t0", "spms/rawid"],
+        ).view_as("ak")
```
gipert: i think in this function we also want a mechanism to apply some filter from outside? for example what if you want to exclude events where the muon flag is true? maybe we can add an argument like

```python
def get_forced_trigger_library(..., where: Callable[[ak.Array], ak.Array] | None = None):
```

that can be used like e.g.

```python
out = get_forced_trigger_library(
    ...,
    where=lambda evt: evt.trigger.muon == False
)
```

workflow/src/legendsimflow/reboost.py
```diff
@@ -558,3 +558,138 @@ def cluster_photoelectrons(
...
+    return ak.Array(
+        {
+            "npe": npe,
+            "t0": t0,
+            "rawid": np.vstack([rawids] * len(npe)),
+        }
+    )
+
+
+def add_forced_triggers(
```
gipert: i am not sure this function is urgent. i think for the moment we want to just have the simulated pes and the rc pes in separate fields. it's very simple to combine them

gipert:
does it make sense to have this function get the rcs from the pulser and forced triggers (full traces) and then move to the first part of the trace of ge triggered events? i.e. it should really try to get everything possible from the data

workflow/src/legendsimflow/reboost.py
rosannadeckert: i encountered this in p13, that i had a 16µs long trace and i wanted two 6µs long windows in the trace. so I shortened the "gap" to 4µs. but maybe for the standard L200 data it's not needed.

gipert: ok we can keep it then, but the library does not only contain independent rc samples by construction. maybe we can add a warning.

in general i think it' cleaner to split the work into a function that gets the largest statistically independent library from data and a function that does the dodgy tricks of increasing the statistics (e.g. using samples multiple times)

rosannadeckert:
we could do that, but we will not add much: consider eg. 3e6 phy events in p16 (without selection, just muon veto). we allow 4 windows in the first half of the trace --> 12e6 forced trigger-like events. of the 3e6 phy events, 3e5 are forced trigger or pulser events. from these you can get 4 additional windows because you can also use the second half of the trace (window in the center 48µs to 55µs needs to be excluded in all cases due to HPGe-SiPM crosstalk or effects of shifted LAr-triggered events in forced trigger data). so you get 4*3e5 =12e5 additional forced trigger-like events --> 10% more in total. is it worth it?

gipert:
well this is for one specific setting, but maybe in some other run (also in the future) the rate of aux events is higher? anyways i think 10% more is worth it

workflow/src/legendsimflow/reboost.py
rosannadeckert: i removed it

rosannadeckert:
Now that I added the `where` argument, you can load the forced trigger library from forced trigger/pulser data and from physics data, using the respective window ranges. I added this example to the docstring

```python
# For forced/pulser data
    rc_data_forced = get_forced_trigger_library(
        evt_files,
        win_range_ns=[(1_000, 44_000), (55_000, 100_000)],
        where=lambda evt: evt.trigger.is_forced | evt.coincident.puls  # codespell:ignore puls
    )

    # For HPGe/LAr triggered data
    rc_data_physics = get_forced_trigger_library(
        evt_files,
        win_range_ns=(1_000, 44_000),
        where=lambda evt: ~evt.trigger.is_forced & ~evt.coincident.puls  # codespell:ignore puls
    )
    
```
then you'd need to concatenate: `rc_data = ak.concatenate([rc_data_forced, rc_data_physics])`

rosannadeckert:
> If you tell me your favourite evt tier file (with enough data for tests to be meaningful), I can upload a truncated version to legend-testdata so you can add some unit tests

in the first 300 events of `prod-blind/auto/latest/generated/tier/evt/ssc/p16/r008/l200-p16-r008-ssc-20251006T205904Z-tier_evt.lh5` there are 2 forced trigger and 2 pulser events

rosannadeckert:
> Now that I added the `where` argument, you can load the forced trigger library from forced trigger/pulser data and from physics data, using the respective window ranges. I added this example to the docstring
> 
> ```python
> # For forced/pulser data
>     rc_data_forced = get_forced_trigger_library(
>         evt_files,
>         win_range_ns=[(1_000, 44_000), (55_000, 100_000)],
>         where=lambda evt: evt.trigger.is_forced | evt.coincident.puls  # codespell:ignore puls
>     )
> 
>     # For HPGe/LAr triggered data
>     rc_data_physics = get_forced_trigger_library(
>         evt_files,
>         win_range_ns=(1_000, 44_000),
>         where=lambda evt: ~evt.trigger.is_forced & ~evt.coincident.puls  # codespell:ignore puls
>     )
>     
> ```
> 
> then you'd need to concatenate: `rc_data = ak.concatenate([rc_data_forced, rc_data_physics])`

i realized this is stupid because you'd have to load the evt tier data twice. So i removed the `where` argument

gipert:
https://github.com/legend-exp/legend-testdata/pull/50

workflow/src/legendsimflow/reboost.py
```diff
@@ -589,3 +589,181 @@ def smear_photoelectrons(
...
+        Default: [(1_000, 44_000), (55_000, 100_000)].
+    ge_trig_range_ns
+        Window ranges for HPGe/LAr trigger events, as list of (start, end) tuples in nanoseconds.
+        Default: [(1_000, 44_000)].
+
+    Returns
+    -------
+    Array with fields "npe", the number of pe per SiPM and per hit,
+    "t0", the time relative to the start of a window in the trace, per SiPM and per hit (makes sure that t0 are between bounds specified in time_domain_ns),
+    and "rawids" the SiPM channel numbers.
```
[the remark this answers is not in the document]
rosannadeckert: i dont want to write "with the same per-hit structure as "npe" and "t0"." because it's not really the same per-hit structure, but rather per-channel for the rawids.

workflow/src/legendsimflow/reboost.py
```diff
@@ -589,3 +589,206 @@ def smear_photoelectrons(
...
+    else:
+        rawid = np.vstack([rawids] * len(npe))
+
+    return ak.Array(
+        {
+            "npe": npe,
+            "t0": t0,
+            "rawid": rawid,
+        }
+    )
```
gipert: minor: i would just return the three arrays as a tuple here

gipert:
looks good to me. next steps:

- implement run partitioning in opt.py (i can do this in a separate pr)
- write a function that returns a list of evt files given a data prod path `l200data` and a runid `runid`. this is a simple function, we have already many similar examples in this repo
- add routines to "augment" the rc sample library in case there aren't enough events

gipert:
after #99 is merged you can rebase and keep going!

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -128,15 +128,24 @@ def process_sipm(
...
+                    ft_offset["idx"] : ft_offset["idx"] + len(chunk)
+                ]
+                ft_offset["idx"] += len(chunk)
+
+                rc_amps, rc_times = spms_pars.forced_trig_sipm_data(
+                    chunk_ft_library, sipm, sipm_uid
+                )
+            else:
+                rc_amps = ak.Array([])
+                rc_times = ak.Array([])
```
gipert: i think this will not work, it needs to be an array of lenght `len(chunk)` filled with empty arrays.

workflow/src/legendsimflow/reboost.py
```diff
@@ -808,7 +810,7 @@ def get_forced_trigger_library(
         rawids = rawids_tmp
 
         # Update counter and check if we've processed enough events
-        num_processed += len(npe)
+        num_processed += len(npe) - npe_len_before
```
gipert: i'm confused about this subtraction, why?

rosannadeckert: yes, you're right. I can just check the len of `npe`, instead of having an extra variable `num_processed`, because `npe` is continously updated in the loop

workflow/src/legendsimflow/scripts/tier/opt.py
rosannadeckert: I ran it with the per-sipm option and it worked, but i'll change it accordingly.

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -167,6 +200,19 @@ def process_sipm(
...
     log.info(msg)
 
+    # Load forced trigger library and pre-sample for this partition
+    msg = "loading forced trigger library for random coincidences"
+    log.debug(msg)
+    with perf_block("load_ft_library()"):
+        evt_files = spms_pars.lookup_evt_files(l200data, runid, evt_tier_name)
+
+        ft_library = reboost_utils.get_forced_trigger_library(
+            evt_files, evt_idx_range[1]
```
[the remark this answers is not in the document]
gipert: @rosannadeckert confirm?

gipert:
I tested this and it seems like loading the library in memory takes quite a while. I would like to see a profile of the code to see if there is any major bottleneck...

gipert:
but very nice!
image

workflow/src/legendsimflow/scripts/tier/opt.py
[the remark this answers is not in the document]
rosannadeckert: yes

rosannadeckert:
the problem is not the file I/O. it's the window processing as I concatenate growing arrays. i'll try to find a solution

gipert:
yeah I thought so

rosannadeckert:
something weird happened to the energy per event but idk if this was my fault. The random coincidences still loog good 

image

gipert:
so we have this max_pe_per_hit threshold set to 10 for the moment, which means that we never see more than 10 pes per event...

rosannadeckert:
oh ok, why? because the convolution with the map is slow?

gipert:
yes, this is an attempt to see if we can speed up the convolution for sources with a lot of energy deposited in lar

workflow/src/legendsimflow/reboost.py
```diff
@@ -654,14 +648,10 @@ def _process_spms_windows(
             npe_list.append(npe_tmp)
             t0_list.append(t0_tmp)
 
-    # Concatenate all results at once
```
gipert: the AI likes very much to suggest to concatenate all at once, does it really speed it up significantly?

rosannadeckert: yes, it does. here's before the change:
```2026-02-21 05:25:35|DEBUG|legendsimflow.reboost @reboost.py:L818 >>> forced-trigger library summary: files_processed=122 total_wall_s=176.8 read_wall_s=7.283 forced_or_pulser_wall_s=84.53 geds_wall_s=84.68 returned_events=912840 requested_min_events=909718
```
and here after
```2026-02-21 06:11:34|DEBUG|legendsimflow.reboost @reboost.py:L816 >>> forced-trigger library summary: files_processed=121 total_wall_s=10.8 read_wall_s=2.508 forced_or_pulser_wall_s=4.571 geds_wall_s=3.439 returned_events=916732 requested_min_events=909718
```

workflow/src/legendsimflow/scripts/plots/tier_opt_observables.py
```diff
@@ -49,35 +49,70 @@ def fig(table):
...
+    plot.plot_hist(h_peamp_ns, ax, n_nans=n_nans(energy), label="non-saturated")
+
+    h_peamp_sim = hist.new.Reg(
         300, 0, 20, name="light per cluster (photoelectrons)"
     ).Double()
-    h_peamp.fill_flattened(energy)
-    plot.plot_hist(h_peamp, ax, n_nans=n_nans(energy), label="non-saturated")
+    h_peamp_sim.fill_flattened(data.energy)
+
+    plot.plot_hist(h_peamp_sim, ax, n_nans=n_nans(data.energy), label="simulated")
```
gipert: i don't understand why you added this, is it a leftover from the rebase?

workflow/src/legendsimflow/scripts/plots/tier_opt_observables.py
```diff
@@ -94,7 +129,23 @@ def fig(table):
 
     ax.set_ylabel("counts")
     ax.legend()
+
+    h_npe_sim = hist.new.Reg(
+        200, 0, 150, name="light per event (photoelectrons)"
+    ).Double()
+    h_npe_sim.fill(ak.sum(data.energy, axis=-1))
+    plot.plot_hist(h_npe_sim, ax, flow="hint", label="simulated")
```
gipert: same comment here, this should be removed

workflow/src/legendsimflow/scripts/plots/tier_opt_observables.py
rosannadeckert: so you only want to plot the non-saturated events with label non-saturated?

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -176,6 +224,29 @@ def process_sipm(
...
+            rc_library = reboost_utils.get_random_coincidences_library(
+                evt_files, n_events_partition
+            )
+            if len(rc_library) == 0:
+                msg = "no random coincidences available, cannot continue"
+                raise RuntimeError(msg)
+    else:
+        rc_library = None
+
+    # Offset tracker for iterating through the pre-sampled library
```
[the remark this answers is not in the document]
gipert: @rosannadeckert this is a good point. i have implemented the first suggestion in the latest commit

PR #94, opened 2026-02-07
Tooling to build realistic pulse shape libraries

SalehGiovanna:
## Summary

Introduces the Python module `legendsimflow.psl` and a driver script to build a **realistic pulse-shape library (PSL)** from an ideal (simulated) one by applying the full electronics response and waveform post-processing chain.

### New Python library (`workflow/src/legendsimflow/psl.py`)

- `build_electronics_response_kernel` — builds a Gaussian × causal-exponential kernel modelling the combined digitizer bandwidth and preamplifier response
- `apply_electronics_response` — batched FFT convolution of waveform arrays with the response kernel (memory-efficient, processes in configurable batches)
- `align_waveforms_to_peak` — shifts each current waveform so its maximum lands at a fixed sample index
- `make_realistic_pulse_shape_lib` — full post-processing chain: unit conversion → convolution → charge-to-current differentiation → moving-window average → drift-time extraction → peak alignment → reshape and NaN-mask restoration

### New driver script (`workflow/src/legendsimflow/scripts/make_hpge_realistic_pulse_shape_lib.py`)

CLI entry point that reads an ideal PSL from an LH5 file, runs the post-processing chain with configurable kernel parameters, and writes the realistic PSL back to HDF5.

### Julia fixes (`workflow/src/LegendSimflow.jl/src/pulse_shape_sim_helpers.jl`)

- `compute_ideal_pulse_shape_lib`: rename internal axis variable `x_axis` → `r_axis` for consistency
- `compute_drift_time_map`: move `T` type parameter to 3rd position to match call sites and `compute_ideal_pulse_shape_lib` convention (previous order caused a silent type mismatch); add missing `nothing` guard after `find_valid_spawn_position`; use `T(NaN)` for the fill value to respect the type parameter

### Other

- Pin `dspeed` to `>=2.1,<2.2` in pixi dependencies
- Add unit tests for all new Python functions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dependencies = [
     "numpy",
     "psutil",
     "pyg4ometry",
+    "dspeed",
     "pygama >=2.3.6",
     "pylegendmeta >=1.3.5",
     "reboost >=0.10.3,<0.11",
@@ -193,6 +194,7 @@ dbetto = ">=1.3.2,<1.4"
 legend-pydataobj = ">=1.17.1"
 legend-pygeom-l200 = ">=0.8,<0.9"
 legend-pygeom-tools = ">=0.2,<0.3"
+dspeed = ">=2.1,<2.2"
 matplotlib = "*"
 numpy = "*"
 pylegendmeta = ">=1.3.5,<1.4"
--- a/workflow/src/legendsimflow/psl.py
+++ b/workflow/src/legendsimflow/psl.py
@@ -203,27 +203,21 @@ def align_waveforms_to_peak(
     # Convert to NumPy for matrix operations
     wfs = ak.to_numpy(wf_input) if isinstance(wf_input, ak.Array) else wf_input
 
-    # Output Array - Indices not covered by the shifted signal are automatically padded with 0.0
     n_wfs, n_samples = wfs.shape
-    shifted_wfs = np.zeros((n_wfs, nsamples_output_current_wfs), dtype=float)
-
-    # Find Amax
     peak_indices = np.argmax(wfs, axis=1)
+    shifts = alignment_idx - peak_indices  # (n_wfs,)
 
-    # Shift each waveform
-    for i in range(n_wfs):
-        peak = peak_indices[i]
-        shift = alignment_idx - peak
+    # For each destination index d, the corresponding source index is d - shift[i].
+    # Build a (n_wfs, n_samples_out) source-index matrix, mask out-of-bounds positions
+    # with 0 (safe to index), then zero out those positions in the result.
+    d = np.arange(nsamples_output_current_wfs)  # (n_samples_out,)
+    src_idx = d[np.newaxis, :] - shifts[:, np.newaxis]  # (n_wfs, n_samples_out)
+    valid = (src_idx >= 0) & (src_idx < n_samples)
+    src_idx_safe = np.clip(src_idx, 0, n_samples - 1)
 
-        # Bounds to prevent IndexErrors
-        start_src = max(0, -shift)
-        end_src = min(n_samples, nsamples_output_current_wfs - shift)
-
-        if start_src < end_src:
-            start_dst = start_src + shift
-            end_dst = end_src + shift
-
-            shifted_wfs[i, start_dst:end_dst] = wfs[i, start_src:end_src]
+    shifted_wfs = np.where(
+        valid, wfs[np.arange(n_wfs)[:, np.newaxis], src_idx_safe], 0.0
+    )
 
     return shifted_wfs, peak_indices
 
--- a/workflow/src/legendsimflow/psl.py
+++ b/workflow/src/legendsimflow/psl.py
@@ -1,0 +1,407 @@
+# Copyright (C) 2026 Luigi Pertoldi <gipert@pm.me>
+#
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Mapping
+
+import awkward as ak
+import numpy as np
+from lgdo import Array, Scalar
+from reboost import units
+from scipy.signal import convolve, fftconvolve
+
+logger = logging.getLogger(__name__)
+
+
+def build_electronics_response_kernel(
+    dt: float,
+    mu_bandwidth: float,
+    sigma_bandwidth: float,
+    tau_rc: float,
+    gaussian_only: bool = False,
+    *,
+    kernel_length: int = 600,
+    kernel_start: int = -100,
+) -> np.ndarray:
+    """Create the system response kernel (gaussian + exponential decay).
+
+    This is obtained by convolving a Gaussian (representing the digitizer
+    bandwidth) with a causal exponential decay (representing the preamplifier
+    response). The kernel is normalized to have a sum of 1.
+
+    Note
+    ----
+    The 'full' mode of convolution results in a length of 2*kernel_length - 1.
+    If `gaussian_only` is True, the kernel will have a length of
+    `kernel_length` and will only contain the Gaussian component, since no
+    convolution is performed.
+
+    Parameters
+    ----------
+    dt
+        The time step between samples in the waveform (in ns)
+    mu_bandwidth
+        The mean of the Gaussian representing the digitizer bandwidth (in ns)
+    sigma_bandwidth
+        The standard deviation of the Gaussian representing the digitizer
+        bandwidth (in ns)
+    tau_rc
+        The time constant of the exponential decay representing the
+        preamplifier response (in ns)
+    gaussian_only
+        If True, only use the Gaussian component (default is False)
+    kernel_length
+        The total length of the response kernel in samples (default is 600)
+    kernel_start
+        The starting index of the kernel relative to the waveform (default is
+        -100, meaning the kernel will cover from -100 to 500 samples)
+
+    Returns
+    -------
+    rf
+        The normalized response kernel
+    """
+
+    # Validate inputs
+    if dt <= 0:
+        msg = f"dt must be positive, got {dt}"
+        raise ValueError(msg)
+    if sigma_bandwidth <= 0:
+        msg = f"sigma_bandwidth must be positive, got {sigma_bandwidth}"
+        raise ValueError(msg)
+    if not gaussian_only and tau_rc <= 0:
+        msg = f"tau_rc must be positive, got {tau_rc}"
+        raise ValueError(msg)
+
+    # Convert to samples
+    mu_samples = mu_bandwidth / dt
+    sigma_samples = sigma_bandwidth / dt
+    tau_samples = tau_rc / dt
+
+    # Define the sample range
+    x = np.arange(kernel_start, kernel_start + kernel_length)
+
+    # Compute Gaussian response (digitizer bandwidth)
+    rf_digi = np.exp(-0.5 * ((x - mu_samples) / sigma_samples) ** 2)
+
+    # Validate before normalization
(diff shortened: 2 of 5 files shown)


tdixon97:
Few suggestions:
- We could try to apply the moving windows with dspeed? (maybe just an enhancement)
- How should the metadata be handled?
- Definitely need some tests

gipert:
I can review this if desired

tdixon97:
@gipert yes that would be great, its almost ready.
I see two main things to fix, adding tests, and integrating with snakemake.

tdixon97:
@gipert we get a test failure totally unrelated to the changes, any ideas?

gipert:
oof, maybe some new pint release?

tdixon97:
I do not think so, the code in l200 should work. I think it is that another package is somehow altering the enviroment and making it not work anymore

tdixon97:
Yes, exactly it fails only after import dspeed

gipert:
i saw already problems related to the pint registry due to dspeed in the past... i don't remember the fix

tdixon97:
I pushed the same fix as you did before, to import dspeed only in local scope

pyproject.toml
```diff
@@ -62,6 +62,7 @@ dependencies = [
     "numpy",
     "psutil",
     "pyg4ometry",
+    "dspeed",
```
gipert, line 65: this also needs to be added to the pixi section with a version bound

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,463 @@
...
+
+        # Slice to original length to maintain timing relative to start. This preserves the initial points to emulate baseline
+        wf_conv_sliced = wf_conv[:, : wf_matrix.shape[1]]
+        convolved_results.append(wf_conv_sliced)
+
+    # Concatenate batches and return as Awkward Array
+    return ak.concatenate(convolved_results)
+
+
+def to_current(wf_array: np.ndarray, dt: float, dt_data: float = 16) -> np.ndarray:
```
gipert: i guess this does not need to be a function (one line), unless you plan to do this with dsPeed soon

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,463 @@
...
+        The type of moving window to apply (see {func}`dspeed.processors.moving_window_multi` for details)
+
+    Returns
+    -------
+    mwa_currents
+        Array of filtered waveforms - same shape as input
+    """
+    from dspeed.processors import moving_window_multi  # noqa: PLC0415
+
+    wf_in = np.ascontiguousarray(wf_array, dtype=float)
```
gipert: why?

tdixon97: @SalehGiovanna ? I think this is probably not needed

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,463 @@
...
+    derivatives
+        2D array containing the current waveforms - same shape as input
+    """
+
+    return (
+        np.diff(wf_array, axis=-1, prepend=0) / dt
+    ) * dt_data  # [current] = 1/data_sample
+
+
+def apply_mwa(
```
gipert: i would maybe have the same comment about this... if you really want to keep the func, i would hide it with a _ in front and remove the docstring

tdixon97: we can add the _ but I don't see an advantage in removing documentation

workflow/src/legendsimflow/psl.py
gipert: otherwise hide with underscore in front and remove docstring

workflow/src/legendsimflow/psl.py
gipert: if the function is very simple there is no need for documentation... let's keep the codebase clean/concise with just what we need

pyproject.toml
tdixon97: I think the version bound is not really needed

workflow/src/legendsimflow/scripts/make_hpge_realistic_pulse_shape_lib.py
gipert, line 1: rename file: wf_map -> psl or whatever you like. decide on a name for "pulse shape library" and use that everywhere

gipert: also this should be a snakemake script to be consisten with the other scripts...

tdixon97: I think that is part of the snakemake integration

workflow/src/legendsimflow/psl.py
gipert: ?

tdixon97: but this is not a simple one line function

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,399 @@
...
+    wf_in = np.ascontiguousarray(wf_array, dtype=float)
+
+    # Preallocate output array
+    mwa_currents = np.zeros_like(wf_in, dtype=float)
+
+    moving_window_multi(wf_in, length, num_mw, mw_type, mwa_currents)
+    return mwa_currents
+
+
+def shift_array(
```
gipert: bad name

workflow/src/legendsimflow/psl.py
SalehGiovanna: True not needed, i think i just needed to make sure it was np and not ak array

workflow/src/legendsimflow/psl.py
gipert: two lines if you exclude import and useless ascontiguousarray, of which one is just a simple array allocation

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,399 @@
...
+
+    # Output Array - Indices not covered by the shifted signal are automatically padded with 0.0
+    n_wfs, n_samples = wfs.shape
+    shifted_wfs = np.zeros((n_wfs, nsamples_output_current_wfs), dtype=float)
+
+    # Find Amax
+    peak_indices = np.argmax(wfs, axis=1)
+
+    # Shift each waveform
+    for i in range(n_wfs):
```
gipert: will this ever be slow?

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,399 @@
...
+        if start_src < end_src:
+            start_dst = start_src + shift
+            end_dst = end_src + shift
+
+            shifted_wfs[i, start_dst:end_dst] = wfs[i, start_src:end_src]
+
+    return shifted_wfs, peak_indices
+
+
+def _validate_pulse_shape_library(
```
gipert: do we really need this?

workflow/src/legendsimflow/psl.py
tdixon97: Yes its maybe not ideal

workflow/src/legendsimflow/psl.py
tdixon97: I think its nice. The function expects an input dict with a very specific format, and so i think its useful to validate this

workflow/src/legendsimflow/psl.py
```diff
@@ -0,0 +1,399 @@
...
+            msg = f"Key '{key}' must be of type Array or Scalar, got {type(pulse_shape_lib[key])}"
+            raise TypeError(msg)
+
+    for item in pulse_shape_lib.values():
+        if not isinstance(item, (Array, Scalar)):
+            msg = f"Key '{key}' must be of type Array or Scalar, got {type(pulse_shape_lib[key])}"
+            raise TypeError(msg)
+
+
+def generate_realistic_map(
```
gipert: map -> psl

workflow/src/legendsimflow/psl.py
SalehGiovanna: time_align_to_fixed_sample ?

workflow/src/legendsimflow/psl.py
tdixon97: I would leave it as it is now and reasses if the code seems slow

workflow/src/legendsimflow/scripts/make_hpge_realistic_pulse_shape_lib.py
gipert: sure i can change that after this pr is merged

gipert:
Claude Code review:
```
  🔴 Critical (must fix before merge)

  1. make_hpge_realistic_pulse_shape_lib.py:71 — spurious mw_pars kwarg passed to
  build_electronics_response_kernel

  build_electronics_response_kernel has no mw_pars parameter — this will crash with a TypeError on every
  run. mw_pars is already correctly passed to make_realistic_pulse_shape_lib on line 80; remove it from
  the kernel call.

  2. psl.py:247–250 — stale key variable in _check_pulse_shape_lib_keys

  for item in pulse_shape_lib.values():      # iterates items as `item`
      ...
      msg = f"Key '{key}' ..."               # `key` is leftover from the previous loop
      raise TypeError(msg)

  When an invalid value is found, the error names whichever key happened to be last in the prior loop
  iteration, not the one that actually failed. Needs .items() to get both.

  3. pulse_shape_sim_helpers.jl vs psl.py:281,348 — axis-order mismatch between Julia and Python

  Julia stores waveforms as [n_samples, n_z, n_r], but the Python docstring and reshape logic
  (reshape(-1, original_shape[-1]), nan_mask = ...any(axis=-1)) both assume [n_z, n_r, n_samples]. This
  will silently corrupt waveform data. The test only uses a 2D array and misses this entirely.

  ---
  🟡 Warning

  4. pulse_shape_sim_helpers.jl:552 — sentinel -1 for skipped positions leaks as -1.0

  drift_times[i] = -1

  This ends up in the float drift-time matrix and is not covered by the NaN mask later — it looks like a
  valid large negative drift time. Should be NaN from the start.

  5. pulse_shape_sim_helpers.jl:536,559 — intersect_op shared across @threads

  Intersect(mintot=0) is constructed once and passed into all threads. If it holds mutable state this is
  a data race. At minimum, document why it's safe or construct one per thread.

  6. pulse_shape_sim_helpers.jl:196–197 — maximum(abs.(diff(...))) picks the wrong grid step

  build_simulation_grid_axis prepends/appends out-of-bounds sentinel points, so the diffs include the
  smaller sentinel gaps. Taking maximum of all diffs returns the interior step (which happens to be
  correct only if the sentinels are smaller), but the intent and implementation don't match. Should
  explicitly take the interior step.

  7. make_hpge_realistic_pulse_shape_lib.py:63 — dt fed to kernel without unit check

  dt = ideal_map_obj["dt"].value   # raw number, assumed ns
  rf_kernel = psl.build_electronics_response_kernel(dt, ...)

  make_realistic_pulse_shape_lib converts dt via units_convfact; the script bypasses this. Works only by
  coincidence if stored units happen to be ns.

  8. tests/test_psl.py:109 — "unit" vs "units" in fixture attrs

  The test uses attrs={"unit": "m"} / attrs={"unit": "ns"}, but production code writes attrs={"units":
  ...} (plural). The unit-conversion path may be silently skipped in tests.

  9. psl.py:375 — deferred import inside a loop body

  from dspeed.processors import moving_window_multi  # inside make_realistic_pulse_shape_lib

  Python caches it, but the import lookup still runs on every call. Move it to the top of the file.

  ---
  🟢 Suggestions

  10. pulse_shape_sim_helpers.jl:347 — error message hardcodes 100 V instead of $threshold

  error("Difference between measured and simulated depletion is larger than 100 V!")

  The check uses threshold correctly but the message doesn't.

  11. pulse_shape_sim_helpers.jl:304 — threshold docstring entry has an extra leading space

  -  threshold: vs - \threshold`:` — will render incorrectly.

  12. psl.py:202–228 — align_waveforms_to_peak inner loop is not vectorized

  Per-waveform Python for loop over potentially thousands of grid points. Fully vectorizable with array
  slicing.

  13. tests/test_psl.py — no 3D test case for make_realistic_pulse_shape_lib

  The test uses a (3, 1000) 2D array. The real input is (n_z, n_r, n_samples). The reshape, NaN-mask, and
   output-shape logic are untested for the actual production shape.

  ---
  Positives

  - NaN-mask preservation for invalid detector pixels is well thought out
  - Kernel normalization has clear validation with informative error messages
  - extend_drift_time_map two-phase update avoids order-dependency bugs
  - compute_drift_time_map transposition is correctly documented
  - Batch processing in apply_electronics_response is a good memory management choice

  ---
  Not ready to merge — the three critical issues (runtime crash, stale variable, axis-order mismatch)
  need to be resolved first.
  ```

PR #95, opened 2026-02-07
Improve Julia tools packaging and add tests

tdixon97:
- [x] setup basic infrastructure (real julia package)
- [x] add CI hooks
- [ ] actually write tests

### DIFF
--- a/workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
+++ b/workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
@@ -180,19 +180,19 @@ end
 
 
 """
-    build_simulation_grid_axis(T, boundary, grid_step)
+    build_simulation_grid_axis(boundary::T, grid_step::T) where {T<:AbstractFloat}
 
 Build a grid axis for detector simulation with extended boundary points.
 
 # Arguments
-- `T`: Numeric type (e.g., Float32)
 - `boundary`: Maximum value of the axis (e.g., detector radius or height in meters)
 - `grid_step`: Grid spacing in meters
 
 # Returns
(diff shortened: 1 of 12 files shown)


workflow/src/legendsimflow/scripts/Project.toml
gipert, line 1: this should be moved to the package now

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
gipert, line 1: this does not need the libjl/ level?

workflow/src/legendsimflow/scripts/make_hpge_drift_time_maps.jl
gipert, line 1: this script should not be moved to the package, it's a script directly used in a snakemake rule and should live with all the other scripts

gipert:
also missing: update snakemake rules

workflow/src/LegendSimflow.jl/src/drift_time_helpers.jl
tdixon97: Not really no

tdixon97:
Yes there is still quite a lot to do.

gipert:
Please also add some constraints to the julia package versions in the Project.toml file.

tdixon97:
@gipert I think we should finish and merge this structural change (tomorrow?) and then add more tests and details later. I am not so sure what changes you want to Project.toml
@aider-chat-bot do you know what he means?

gipert:
It's very easy and also important to do asap

https://pkgdocs.julialang.org/v1/compatibility/

PR #98, opened 2026-02-10
fix to allow surface sampling to run

tdixon97:
This is not optimal for most sources, but will atleast run without problems in most cases

### DIFF
--- a/workflow/src/legendsimflow/commands.py
+++ b/workflow/src/legendsimflow/commands.py
@@ -311,6 +311,7 @@ def make_remage_macro(
                 ]
                 if sim_cfg.confinement.startswith("~volumes.surface:"):
                     confinement += ["/RMG/Generator/Confinement/SampleOnSurface true"]
+                    confinement += ["/RMG/Generator/Confinement/SurfaceSampleMaxIntersections 1000"]
 
         elif isinstance(sim_cfg.confinement, list | tuple):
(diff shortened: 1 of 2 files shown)


workflow/src/legendsimflow/commands.py
```diff
@@ -311,6 +311,7 @@ def make_remage_macro(
                 ]
                 if sim_cfg.confinement.startswith("~volumes.surface:"):
                     confinement += ["/RMG/Generator/Confinement/SampleOnSurface true"]
+                    confinement += ["/RMG/Generator/Confinement/SurfaceSampleMaxIntersections 100"]
```
gipert: you have to add this also below. maybe deduplicate the code?

tdixon97: huh

tdixon97:
Strangely it doesnt seem to work

tdixon97:
```
[Error ---> Exceeded maximum number of allowed iterations (100), check that your surfaces are efficiently sampleable and try, eventually, to increase the threshold through the de\
dicated macro command /RMG/Generator/Confinement/MaxSamplingTrials. Returning dummy vertex
```

Seems a bug, the max-sampling trials should be higher, but then I do not get why this isnt caught by remage tests

*Edit* This was just a mistake in which template was used.

gipert (approved):

PR #103, opened 2026-02-12
hit: add profiling and speedup

gipert:


gipert:
short hit jobs are now faster with channel usability caching

gipert:
have to cache this in a snakemake rule, otherwise starting the workflow takes quite long

PR #115, opened 2026-02-24
A/E classifier

gipert:


workflow/src/legendsimflow/hpge_pars.py
```diff
@@ -652,6 +726,14 @@ def build_energy_res_func(function: str) -> Callable:
     raise NotImplementedError
 
 
+def build_aoe_res_func(function: str) -> Callable:
+    """A/E resolution function builder."""
+    if function == "SigmaFit":
+        return lambda energy, a, b, c: (a + (b / (energy + 10**-99)) ** c) ** (0.5)
```
tdixon97, line 727: Is it needed to harcode this? Can you not parse the pars file? Or can you atleast assert it has the right format?
My worry is George changing something

gipert: literally now https://github.com/legend-exp/pygama/pull/656

gipert: i will open an issue about this

PR #124, opened 2026-03-17
Update forced trigger library load logic

### LINKED ISSUE #50: Random coincidences correction to photoelectron times

### LINKED ISSUE #114: Improve memory footprint of getting random coincidences in opt
When the partitions are big (e.g. with p16 simulations where 1E8 events end up in a single partition), the random coincidence library (and the making of) can use a lot of memory. Need to implement a smarter solution

rosannadeckert:
Loading the forced trigger library at once, then taking slices of it, uses too much memory for a large number of simulated events. This is why we changed the load logic to load a forced trigger library _per chunk_. 

Generally, we loop over `evt` files. There can be two cases:

**Case 1:** One file is not enough for the number of events in a chunk. Then
- Loops over files until enough events in forced trigger library

**Case 2:** One file produces a forced trigger library that is too large for the required number of events in a chunk. Then:
- residual forced trigger library is held in memory and used for the next chunk
- once the forced trigger library is exhausted, we load the forced trigger library from the next file

If the available event files in data do not provide enough forced trigger events, files are reused.

Performance-wise, for p16 ssc simulation of one job with optical maps per channels, we have:

- with `add_random_coincidences=True`: 
```
s	h:m:s	max_rss	max_vms	max_uss	max_pss	io_in	io_out	mean_load	cpu_time
1973.39	0:32:53	2602.66	5377.70	2497.18	2566.74	0.28	4924.43	98.66	1947.62
```
- with `add_random_coincidences=False`: 
```
s	h:m:s	max_rss	max_vms	max_uss	max_pss	io_in	io_out	mean_load	cpu_time
1017.77	0:16:57	2323.82	5790.32	2230.13	2299.71	7.48	4848.05	97.55	993.46
```

--> memory is fine, processing time doubled...

The problem is that now, the same forced trigger libraries must be loaded for each channel, while before, each channel used a preloaded forced trigger library. This is because of the logic of looping over channels, and then over chunks. Looping over channels makes more sense because you don't want to store the optical maps per channel in memory...

### DIFF
--- a/workflow/src/legendsimflow/scripts/tier/opt.py
+++ b/workflow/src/legendsimflow/scripts/tier/opt.py
@@ -85,11 +85,13 @@ def process_sipm(
     optmap_lar: str | Path | OptmapForConvolve,
     sipm: str,
     sipm_uid: int,
+    evt_tier_name: str | None,
+    hit_tier_name: str | None,
     out_file: str | Path,
     runid: str,
     usability: str,
-    rc_library: ak.Array | None,
-    rc_offset: dict,
+    rc_file_state: dict,
+    rc_index_lookup: dict[str, dict[str, np.ndarray]] | None = None,
 ) -> None:
     with perf_block("load_optmap()"):
         if not isinstance(optmap_lar, OptmapForConvolve):
@@ -145,17 +147,29 @@ def process_sipm(
             pe_times = pe_times_micro
             pe_amps = pe_amps_micro
 
-        # Add random coincidences from forced trigger library
-        if rc_library is not None:
+        # Add random coincidences from forced trigger files
+        if rc_index_lookup is not None:
             with perf_block("add_random_coincidences()"):
                 if usability == "on":
-                    chunk_rc_library = spms_pars.get_rc_library_chunk(
-                        rc_library, len(chunk), rc_offset
+                    if evt_tier_name is None:
+                        msg = "evt_tier_name must be set when random coincidences are enabled"
+                        raise RuntimeError(msg)
+                    if hit_tier_name is None:
+                        msg = "hit_tier_name must be set when random coincidences are enabled"
+                        raise RuntimeError(msg)
+                    rc_evt_files = [str(f) for f in rc_index_lookup]
+                    chunk_rc = spms_pars.get_chunk_rc_data(
+                        rc_evt_files,
+                        rc_file_state,
+                        len(chunk),
+                        evt_tier_name,
+                        hit_tier_name,
+                        sipm,
+                        sipm_uid,
+                        rc_index_lookup,
                     )
-
-                    rc_amps, rc_times = spms_pars.get_sipm_rc_data(
-                        chunk_rc_library, sipm, sipm_uid
-                    )
+                    rc_amps = chunk_rc.npe
+                    rc_times = chunk_rc.t0
                 else:
                     rc_amps = _ak_array_of_empty_arrays(len(chunk))
                     rc_times = _ak_array_of_empty_arrays(len(chunk))
@@ -169,7 +183,7 @@ def process_sipm(
             out_table.add_field("energy", VectorOfVectors(pe_amps))
             out_table.add_field("is_saturated", Array(is_saturated))
 
-            if rc_library is not None:
+            if rc_index_lookup is not None:
                 # TODO: units should be automatically forwarded
                 out_table.add_field(
                     "rc_time", VectorOfVectors(rc_times, attrs={"units": "ns"})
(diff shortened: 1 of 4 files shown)


tdixon97:
I do not get how you can loop over channels, the FT correction is inherently evt based so should be done at that step (not opt).
I think we should first just implement a correction on the summed pe / multiplicity ... which should be less computationally demanding.
I think the simflow should compute what we use *now* for the background modelling / classifer (ie just light / no light) and can be adjusted in future

gipert:
as far as i understand (@rosannadeckert correct me if i'm wrong) this pr is almost done and it does not slow down things. @rosannadeckert will finish it up (including benchmarking) in the next days and then we can review. let's see what the status is before saying that we want something else

rosannadeckert:
it does not increase memory as much anymore, but it does slow down things by a lot (see description). i am loading the evt files for each channel atm, so i will have to change this to loading the hit files with tcm indices. so it's not ready yet, computation time has to go down

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -145,12 +175,96 @@ def process_sipm(
...
+                        )
+
+                    while (
+                        total_rc_events < len(chunk)
+                        and empty_parts_streak < max_empty_parts
+                    ):
+                        rc_evt_file = _next_rc_evt_file(rc_evt_files, rc_file_state)
+                        n_missing = len(chunk) - total_rc_events
+                        part = spms_pars.get_rc_library([rc_evt_file], n_missing)
+                        if len(part) == 0:
```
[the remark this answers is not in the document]
tdixon97: This may well be why the function is slow..

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -145,12 +175,96 @@ def process_sipm(
             pe_times = pe_times_micro
             pe_amps = pe_amps_micro
 
-        # Add random coincidences from forced trigger library
-        if rc_library is not None:
+        # Add random coincidences from forced trigger files
+        if rc_evt_files is not None:
             with perf_block("add_random_coincidences()"):
```
tdixon97, line 152: I think this should definetly be done at `evt` not `opt` (which is "hit" level). This would also improve the performance issue

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -145,12 +175,96 @@ def process_sipm(
...
                 if usability == "on":
-                    chunk_rc_library = spms_pars.get_rc_library_chunk(
-                        rc_library, len(chunk), rc_offset
+                    # Build a per-chunk RC library, adding more evt files if
```
tdixon97: there is alot of pretty complicated code here, is it really needed? If it is, it definitely has to go in the python module not directly in the script

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -80,6 +80,36 @@ def _ak_array_of_empty_arrays(n):
...
 
+def _next_rc_evt_file(evt_files: list[str | Path], rc_file_state: dict) -> str | Path:
+    """Return the next `evt` file, cycling through files in input order before repeating."""
+    if "order" not in rc_file_state:
+        order = list(evt_files)
+        rc_file_state["order"] = order
+        rc_file_state["idx"] = 0
+        rc_file_state["counts"] = {}
+
+    order = rc_file_state["order"]
```
tdixon97: I do not understand what this rc_file_state is needed for

workflow/src/legendsimflow/scripts/tier/opt.py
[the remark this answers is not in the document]
rosannadeckert: I was not done with this pr....

workflow/src/legendsimflow/scripts/tier/opt.py
```diff
@@ -291,8 +400,8 @@ def _make_iterator(det_name=det_name, i_start=i_start, n_entries=n_entries):
                     hit_file,
                     runid,
                     usability,
-                    rc_library,
-                    rc_offset.setdefault(sipm, {"idx": 0}),
+                    rc_evt_files,
+                    {},
```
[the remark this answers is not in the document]
rosannadeckert: no, has to be reset for each channel

workflow/src/legendsimflow/scripts/tier/opt.py
rosannadeckert: we do this at the `opt` tier because of partitioning

workflow/src/legendsimflow/scripts/tier/opt.py
rosannadeckert: yes i agree

workflow/src/legendsimflow/scripts/tier/opt.py
rosannadeckert: `get_rc_library` loads the RC data from one file, if the current chunk needs less events than loaded in the file RC library, the remaining events can be used for the next chunk. `rc_file_state` tracks this

rosannadeckert:
By loading `tcm` and `hit` files instead of `evt` files for the random coincidence correction per channel, i could bring down the time to
```
s	h:m:s	max_rss	max_vms	max_uss	max_pss	io_in	io_out	mean_load	cpu_time
1408.48	0:23:28	2314.14	4929.18	2084.77	2154.32	0.81	4920.96	96.19	1355.27
```
for per-channel optical map with random coincidence correction. Compared to 
```
s	h:m:s	max_rss	max_vms	max_uss	max_pss	io_in	io_out	mean_load	cpu_time
1017.77	0:16:57	2323.82	5790.32	2230.13	2299.71	7.48	4848.05	97.55	993.46
```
without the correction. The majority of the wall time (~70%) is spent on the `tcm`-read block. The `hit`-read block takes only ~5% of wall time.

workflow/src/legendsimflow/scripts/tier/opt.py
tdixon97: Hmm im not sure, its an inherently evt level correction

workflow/src/legendsimflow/spms_pars.py
```diff
@@ -51,8 +49,172 @@ def lookup_evt_files(l200data: str, runid: str, evt_tier_name: str) -> list[str
...
+
+def get_chunk_rc_data(
+    rc_evt_files: list[str | Path],
+    rc_file_state: dict,
+    chunk_size: int,
+    hit_tier_name: str,
+    sipm: str,
+    sipm_uid: int,
+    rc_index_lookup: dict[str, dict[str, np.ndarray]],
+) -> ak.Array:
```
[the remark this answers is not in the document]
rosannadeckert: `rc_evt_files` will not be empty. If empty, error is raised:
```
rc_evt_files = sorted(
                spms_pars.lookup_evt_files(l200data, runid, evt_tier_name)
            )
if len(rc_evt_files) == 0:
    msg = "no random coincidences available, cannot continue"
    raise RuntimeError(msg)
```

workflow/src/legendsimflow/spms_pars.py
```diff
@@ -173,216 +367,143 @@ def get_rc_library(
...
-            "forced-trigger library file %s: forced_or_pulser_events=%d "
-            "geds_events=%d cumulative_events=%d",
-            Path(file).name,
-            n_forced_pulser,
-            n_geds,
-            n_collected_events,
-        )
+    else:
+        tcm_file = Path(str(evt_file).replace("evt", "tcm"))
+        hit_file = Path(str(evt_file).replace("evt", hit_tier_name))
```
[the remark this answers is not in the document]
rosannadeckert: it is true that `evt_tier_name` is not necessarily `evt`. i changed that. but otherwise, the path of `tcm` and `hit` should be the same as `evt`, correct? or should i make it more generic?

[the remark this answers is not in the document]
gipert: it's correct, but the ai is telling you: what if the tiers live in different prod cycles? you need to use the path to the tier specified in the dataflow config and then replace the tier name

workflow/src/legendsimflow/scripts/tier/opt.py
gipert: it is an event correction, but let's first see if this approach works before changing design

gipert: if the performance is bad and we expect it to improve if we move it to evt, we can do that (maybe the AI can too)

gipert:
I have included these commits in #155

PR #144, opened 2026-03-24
refactor: extract par step from hit.smk, rename make_tiers to make_steps

### LINKED ISSUE #80: Tricky to not rerun dt maps
Currently it is not easy to tell snakemake to not rerun drift time maps (unlike for example stp or hit). I think this would be a useful feature

### LINKED ISSUE #123: Do not run some parts
Currently the user is obliged to generate both PSD and LAr sims, for some applications this either might not make sense or the inputs might not exist.
I think we should make these things optional / disableable? @gipert

gipert:
Move all parameter-building rules (simstat partitioning, HPGe drift time
maps, current pulse models, observable models) from hit.smk into a
dedicated par.smk step inserted between stp and opt in STEPS_ORDERED.

Rename the config field make_tiers -> make_steps and the corresponding
TIERS_ORDERED constant to STEPS_ORDERED to reflect that par is not a
simulation output tier. Fix the cross-file rule reference
rules.make_simstat_partition_file.output in build_tier_hit with a direct
path (par.smk is conditionally included). Raise SimflowConfigError when
vtx or par are used as simlist prefixes, since neither produces
simid-scoped outputs.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


docs/source/manual/prod.md
```diff
@@ -61,6 +61,58 @@ stp.fibers_Rn222_to_Po214                                33:20:00    100
...
+run-statistics partitioning files). If `opt` or `hit` is included in
+`make_steps` without `par`, those parameter files **must already exist on
+disk**. Omitting `par` is only safe when the parameters have been produced in a
+previous run and have not changed.
+
+:::
+
+:::{note}
+
+`vtx` and `par` cannot be used as step prefixes in `simlist` items (e.g.
```
tdixon97, line 111: is that true for vtx?

docs/source/manual/setup.md
```diff
@@ -28,9 +28,25 @@ Here's a basic description of its fields:
...
+  Snakemake rule files for the listed steps are loaded, so Snakemake will treat
+  outputs of excluded steps as pre-existing source files rather than targets to
+  (re)build. This is the primary mechanism for protecting existing production
+  outputs from accidental re-execution and for speeding up DAG generation.
+  Available steps (in canonical order): `vtx`, `stp`, `par`, `opt`, `hit`,
+  `evt`, `cvt`. If omitted, all steps are included. The order in the list does
+  not matter — steps are always sorted canonically. A few caveats apply:
+  - Steps `vtx` and `par` do not produce simid-scoped outputs and cannot appear
+    as prefixes in `simlist` items.
+  - The `par` step builds the parameters consumed by both `opt` and `hit` (HPGe
```
tdixon97, line 41: worth to spell out more clearly that its for parameters?

tests/test_aggregate.py
```diff
@@ -50,13 +52,11 @@ def test_process_simlist_is_cumulative(config):
     simid = "birds_nest_K40"
     targets_evt = agg.process_simlist(config, simlist=[f"evt.{simid}"])
 
-    targets_vtx = agg.process_simlist(config, simlist=[f"vtx.{simid}"])
```
tdixon97, line 53: why?

workflow/rules/par.smk
```diff
@@ -0,0 +1,335 @@
...
+    message:
+        "Extracting HPGe observables models for {wildcards.runid}"
+    # NOTE: we don't list the file dependencies here because they are
+    # dynamically generated, and that would slow down the DAG generation
+    output:
+        eresmod_file=patterns.output_eresmod_filename(config),
+        aoeresmod_file=patterns.output_aoeresmod_filename(config),
+        psdcuts_file=patterns.output_psdcuts_filename(config),
+    run:
+        import dbetto
```
tdixon97, line 289: move the following code to a function? It would also be nice if these functions were tested

tdixon97 (approved):
Looks good but I am a bit confused on the treatment of vtx

docs/source/manual/prod.md
gipert: yes it's true, I haven't written an aggregate method for it

tests/test_aggregate.py
gipert: because it didn't return a correct simlist. unfortunately the aggregate rule does not consider that some sims do not vtx yet. it can be implemented in the future

workflow/rules/par.smk
gipert: the reason i didn't do that earlier is that this function would have many arguments and that is a bit ugly...

gipert: at some point soon we should try to configure these tests for the full workflow

docs/source/manual/setup.md
gipert: I don't understand, it is already written?

PR #149, opened 2026-03-25
ci: run dummy production up to stp

gipert:


gipert:
to fix this we need to run these tests via pixi. we do this already in other places like legend-pygeom-hades or legend-pygeom-l200, so we need to copy from there

tdixon97:
I will try to fix it

tdixon97:
Still some work to do, eg. some private metadata is needed. Maybe we can just make a small fake subset

gipert:
we should, that's essentially the task... the best would be to just take what we have in legend-testdata and update it to make it usable here.

gipert:
this is working now. we should slowly enable the higher tiers while implementing pars fallbacks for when the l200data is not available

Issue #127, opened 2026-03-18
Update for latest SSD


tdixon97:
Currently we do not support the latest SSD/ LegendDataManagment versions, @SalehGiovanna already has the code to fix this.

gipert:
#19 was tracking this, let's close one of the two issues.

I hope @hervasa2 can finally dedicate some time now to this issue

hervasa2:
PD im _finally_ on it as we speak

hervasa2:
Waiting on a release from LegendDataManagement from @theHenks

gipert:
Please coordinate with @SalehGiovanna, she is also working on the Julia scripts in #57

gipert:
I would first like to see working scripts with SSD v0.10 with some tests, then we can upgrade to v0.11

tdixon97:
I totally agree @gipert

theHenks:
Release is out: https://github.com/legend-exp/LegendDataManagement.jl/releases/tag/v0.6.3

PR #168, opened 2026-03-30
Upgrade to SSD-0.11

hervasa2:
This PR addresses #127
Changed `[compat]` and added minor changes to support and require SSD > 0.11. 
 ```julia
charge_drift_model = ADLChargeDriftModel(
        "$meta_path/simprod/config/pars/geds/ssd/adl-2016-temp-model.yaml", T = T
    )
```
Is no longer needed as LegendDataManagement + SSD now supports temperature dependence, with no additional config files. The support is now completely handled in SSD with the model loaded by LegendDataManagement.

I am leaving `adl-2016-temp-model.yaml` in there for now. But we should consider deleting once/if we run any tests to compare. 
Added tests to check that new model is loaded correctly.

### DIFF
--- a/workflow/rules/stp.smk
+++ b/workflow/rules/stp.smk
@@ -28,6 +28,10 @@ rule gen_all_tier_stp:
         aggregate.gen_list_of_all_simid_outputs(config, tier="stp"),
 
 
+# NOTE: the geometry rules below intentionally live in stp.smk rather than a
+# shared module. The geometry must always be produced as part of an stp tier
+# production — upper tiers (vtx, hit, opt, ...) are required to take the
+# geometry from an existing stp production and must never build their own.
 rule gen_geom_config:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: workflow/src/LegendSimflow.jl/Project.toml [hunk 1]
```diff
@@ -21,7 +21,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 test = ["Test"]
 
 [compat]
-LegendDataManagement = "0.6"
+LegendDataManagement = ">=0.6.3"
```
gipert: i would also add an upper bound here, i don't want things to break at some point

tdixon97:
Looks good to me

gipert:
@hervasa2 nice thank you! how much is this tested?

hervasa2:
On the test detector all the way to the highest level functions in the Julia environment.

PR #185, opened 2026-04-03
feat: add PSD usability support to hit tier

### LINKED ISSUE #183: PSD usability
Need to add this as an additional column to the hit tier.

gipert:
## Summary

- Adds `PSD_USABILITY_CODE` encoding (`valid=0`, `present=1`, `missing=2`) and `encode_psd_usability`/`decode_psd_usability` to `metadata.py`, mirroring the existing `USABILITY_CODE` pattern
- Extends `gen_list_of_all_usabilities()` in `aggregate.py` to return `{usability, psd_usability}` per detector, reading `psd.status.low_aoe` from the channel map status metadata
- Writes `psd_usability` as an integer column in the `hit` tier output
- Adds warnings in `hit.py` and `opt.py` when usability metadata is absent for a detector
- Updates tests to cover the new encoding and the warning/fallback path for unknown `psd.status.low_aoe` values

## Test plan

- [x] `pixi run test-python` passes
- [x] `pre-commit run --all-files` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/AGENTS.md
+++ b/tests/AGENTS.md
@@ -8,3 +8,14 @@ Python tests are stored in `tests/` and managed with Pytest. Julia tests are in
 - `test_workflow.py`: integration Snakemake testing of the workflow with a dummy
   production (configured in `tests/dummyprod`) that can be tested in CI
 - `l200data/`: test data for the LEGEND-200 data production
+
+## Test data construction
+
+`conftest.py` checks out a pinned commit of `legend-testdata` and copies the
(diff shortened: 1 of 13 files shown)


tdixon97:
I think it should be optional, there have been times when its reliable

gipert:
huh?

gipert:
this only adds it as a flag, it does not disable simulating psd

PR #186, opened 2026-04-03
Optimize memory usage in noise waveform extraction and plotting

### LINKED ISSUE #181: `extract_hpge_current_pulse_model.py` is using too much memory
I've seen single jobs using up to 4GB, we should investigate.

gipert:
## Summary
This PR improves memory efficiency in the HPGe parameter extraction workflow by optimizing waveform buffer management and adding explicit resource cleanup in plotting operations.

### `hpge_pars.py` - `get_noise_waveforms()` function
- **Pre-allocated buffer instead of list accumulation**: Replaced the inefficient pattern of appending small arrays to a list with a pre-allocated NumPy array that grows dynamically when needed
  - Initial capacity set to `maximum_number` (if specified) or 10,000 samples
  - Array doubles in size when capacity is exceeded, reducing reallocation overhead
- **Improved indexing**: Changed `indices[0]` iteration to direct iteration over `indices` after extracting with `[0]` indexing
- **Memory cleanup**: Added explicit `del` statements for intermediate arrays (`energies`, `indices`) to free memory during iteration
- **Simplified return logic**: Returns properly-sized slice of buffer (`buf[:count]`) instead of concatenating multiple arrays

### `extract_hpge_current_pulse_model.py` - Resource management
- **Added explicit figure cleanup**: Added `plt.close(fig)` calls after saving figures to PDF to prevent memory accumulation
- **Added variable cleanup**: Added `del` statements for large intermediate arrays after use:
  - `all_dts`, `selected_dts` after drift-time selection plot
  - `times_list`, `current_list` after current model plot
  - `noise_wfs`, `temp` after noise waveform analysis
- **Reduced maximum waveforms**: Lowered `maximum_number` from 100,000 to 10,000 in noise waveform extraction for more reasonable memory footprint
- **Added matplotlib import**: Imported `pyplot` explicitly for figure management

## Implementation Details
The waveform buffer optimization uses a doubling strategy for dynamic array growth, which provides O(1) amortized insertion time while avoiding the overhead of list-to-array conversion. The explicit resource cleanup prevents memory leaks in long-running batch processing workflows.

https://claude.ai/code/session_019YYgvbfRTrHjx34X89Az8q

### DIFF
--- a/workflow/src/legendsimflow/hpge_pars.py
+++ b/workflow/src/legendsimflow/hpge_pars.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import functools
+import itertools
 import logging
 import re
 from collections.abc import Callable
@@ -469,7 +470,12 @@ def get_current_pulses(
     return times_list, current_list
 
 
-def get_noise_waveforms(
+def _remove_outliers(data: NDArray, sigma: float = 5) -> NDArray:
+    """Remove elements more than ``sigma`` standard deviations from the mean."""
+    return data[abs(data - np.mean(data)) < sigma * np.std(data)]
+
+
+def _iter_noise_waveforms(
     raw_files: list,
     hit_files: list,
     lh5_group: str,
@@ -478,14 +484,51 @@ def get_noise_waveforms(
     *,
     threshold: float = 5,
     length: int = 1000,
+    energy_var: str = "cuspEmax_cal",
+):
+    """Yield noise waveforms one at a time without accumulating them all in memory.
+
+    Parameters are the same as :func:`get_noise_maxima_and_sample`.
+    """
+    from dspeed.vis import WaveformBrowser  # noqa: PLC0415
+
+    for raw_file, hit_file in zip(raw_files, hit_files, strict=True):
+        browser = WaveformBrowser(
+            str(raw_file), lh5_group, dsp_config=dsp_config, lines=[dsp_output]
+        )
+        energies = lh5.read(
+            f"{lh5_group.replace('raw', 'hit')}/{energy_var}", hit_file
+        ).view_as("np")
+
+        indices = np.where(energies < threshold)[0]
+        del energies
+
+        for idx in indices:
+            browser.find_entry(idx, append=False)
+            yield browser.lines[dsp_output][0].get_ydata()[:length].copy()
+
+        del indices
+
+
+def get_noise_maxima_and_sample(
+    raw_files: list,
+    hit_files: list,
+    lh5_group: str,
+    dsp_config: str,
+    dsp_output: str,
+    template: ArrayLike,
+    *,
+    norm: float = 1,
+    sample_size: int = 100,
+    threshold: float = 5,
     maximum_number: int | None = None,
     energy_var: str = "cuspEmax_cal",
-) -> NDArray:
-    """Extract a matrix of noise waveforms, after applying some DSP processing.
+) -> tuple[NDArray, NDArray]:
+    """Compute waveform maxima on-the-fly, keeping only a small sample in memory.
 
-    The waveforms are only those with energy less than `threshold` and are the
-    result of the dsp processing defined in `config_path` with output variable
-    `dsp_output`.
+    This avoids storing all noise waveforms at once. Instead, it iterates
+    through waveforms, computes the maximum of ``waveform + template`` for each,
+    and only retains the first ``sample_size`` waveforms for plotting.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


workflow/src/legendsimflow/scripts/extract_hpge_current_pulse_model.py
```diff
@@ -123,23 +128,26 @@
         str(dsp_cfg_file),
         "curr_av",
         length=len(temp),
-        maximum_number=100000,
+        maximum_number=10000,
```
gipert: @tdixon97 not sure but maybe the problem was the open plots

tdixon97: I am not sure changing this is a good idea

tdixon97: I doubt the plots use much memory..

gipert: Reverted the `maximum_number` change back to 100000 in 5ecd03d — agreed it's better to keep the original value.

The remaining changes that should still help with memory:
- **Pre-allocated array** in `get_noise_waveforms()`: avoids accumulating a list of 100k small arrays and then `np.concatenate()`, which temporarily doubles memory (~2.4GB peak → ~1.2GB for the waveform buffer alone)
- **`plt.close(fig)`** after each PDF save: matplotlib does hold figure state in memory, and with multiple figures open simultaneously it adds up (though likely not the dominant factor)
- **`del` for large intermediates** (`noise_wfs`, `energies`, etc.): frees memory earlier rather than waiting for scope exit

---
_Generated by [Claude Code](https://claude.ai/code)_

tdixon97: If the noise waveforms use alot of memory we should change a bit the design and instead iterate storing only say 100 waveforms, and for the rest just store A_max

tdixon97: I think the AI could manage it very easily

tdixon97: Another way to save memory would be to store shorter nosie waveforms, you probably only need ~200 ns close to A_max

gipert: Implemented the on-the-fly approach in be4aa0d. The new `get_noise_maxima_and_sample()` function iterates through waveforms via a generator, computes `max(wf + template)` for each one immediately, and only keeps the first 100 waveforms in memory for the diagnostic plot. Peak memory drops from ~1.2GB to ~1.2MB (plus the 1D maxima array).

Regarding shorter noise waveforms around A_max — that's a good idea for a further optimization but would require changes to the DSP config / waveform extraction logic to window around the peak. Happy to look into it if you'd like, but it could also be a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

tdixon97: how is it so fast

gipert: claude is reading you, can you tell it what to do precisely?

gipert: cluade is good

workflow/src/legendsimflow/hpge_pars.py
```diff
@@ -469,6 +469,51 @@ def get_current_pulses(
...
+            yield waveform[:length].copy()
+            count += 1
+
+            if maximum_number is not None and count >= maximum_number:
+                return
+
+        del indices
+
+
 def get_noise_waveforms(
```
tdixon97: What is the value in such a simple function? Is it used somewhere? Or is it just kept for backwards compatibility?

tdixon97:
Looks great

workflow/src/legendsimflow/hpge_pars.py
gipert: Good call — removed it in 8a1a02a. It was only used by the test, which now tests `_iter_noise_waveforms` directly.

---
_Generated by [Claude Code](https://claude.ai/code)_

workflow/src/legendsimflow/hpge_pars.py
```diff
@@ -500,45 +547,58 @@ def get_noise_waveforms(
...
-            waveforms.append(waveform[:length].reshape(1, length))
+    for wf in _iter_noise_waveforms(
+        raw_files,
+        hit_files,
+        lh5_group,
+        dsp_config,
+        dsp_output,
+        threshold=threshold,
+        length=length,
+        maximum_number=maximum_number,
```
tdixon97: Should this be called with a smaller maximum_number to reduce the memory usage?

gipert: With the on-the-fly approach, `maximum_number=100000` is no longer memory-heavy — waveforms are processed one at a time via the generator and immediately discarded. The only things accumulated are:
- `maxima`: a list of 100k floats (~800KB)
- `sample_wfs`: capped at `sample_size=100` waveforms (~1.2MB)

So 100k is fine to keep here for statistical accuracy without the memory cost.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #191, opened 2026-04-03
Average waveform construction

SalehGiovanna:
Workflow proposal for the SSC superpulse construction. 

- [x]  Define API, type hints, and docstrings for dataset preparation (EVT/DSP/raw merging, pixel slicing).

- [x] Define API, type hints, and docstrings for superpulse construction (preprocessing, chi2 cuts, averaging).

- [x] Implement the internal logic of functions:
1. Steps needing particular review are marked with ### comments
2. Now that the code is there, docstrings can be significantly reduced for most of the functions
3. get_charge_and_current_wfs_for_slice function requires specific checks: it currently works (does not break in the tests I made), but the output is not exactly the same I had in the original code. Should decide which are the best charge/current waveforms to be extracted from processing chain, and how to handle them after extraction (Upsampled charge too? Trim superpulses to same length across all slices? Fixed length? Which one?

- [x] Add tests to verify outputs match the existing notebook implementation.

### DIFF
--- a/tests/test_superpulses.py
+++ b/tests/test_superpulses.py
@@ -1,30 +1,9 @@
-"""Tests for legendsimflow.superpulses.
-
-Covered functions
------------------
-- Slice
-- Superpulse.__init__
-- Superpulse.__repr__
-- Superpulse.to_lgdo
-- perform_data_selection
-- select_detector_events
-- select_data_in_slice
-- compute_superpulse
-- compute_chi2_vs_superpulse
-- apply_chi2_cut
-- write_superpulses_to_lh5
-
-"""
-
 from __future__ import annotations
 
 import awkward as ak
 import numpy as np
 import pytest
-from lgdo import Array as LGDOArray
-from lgdo import Scalar as LGDOScalar
-from lgdo import Struct as LGDOStruct
-from lgdo import lh5
+from lgdo import Array, Scalar, Struct, lh5
 
 from legendsimflow.superpulses import (
     Slice,
@@ -32,7 +11,6 @@ from legendsimflow.superpulses import (
     apply_chi2_cut,
     compute_chi2_vs_superpulse,
     compute_superpulse,
-    perform_data_selection,
     select_data_in_slice,
     select_detector_events,
     write_superpulses_to_lh5,
@@ -68,7 +46,7 @@ def _make_superpulse(
 
 def _make_evt_data(
     n_events=20,
-    rawids=None,
+    detector_names=None,
     hit_idxs=None,
     multiplicity=1,
     forced=False,
@@ -89,8 +67,8 @@ def _make_evt_data(
     All per-detector arrays are wrapped in a length-1 inner dimension to
     reproduce the multiplicity==1 guarantee of the real EVT tier.
     """
-    if rawids is None:
-        rawids = [[1084803]] * n_events
+    if detector_names is None:
+        detector_names = [["V03422A"]] * n_events
     if hit_idxs is None:
         hit_idxs = [[i] for i in range(n_events)]
 
@@ -103,7 +81,7 @@ def _make_evt_data(
                 "muon_offline": [muon_offline] * n_events,
             },
             "geds": {
-                "rawid": rawids,
+                "detector_name": detector_names,
                 "hit_idx": hit_idxs,
                 "multiplicity": [multiplicity] * n_events,
                 "quality": {
@@ -230,7 +208,7 @@ def test_superpulse_repr():
 def test_to_lgdo_returns_struct():
     sp = _make_superpulse()
     result = sp.to_lgdo()
-    assert isinstance(result, LGDOStruct)
+    assert isinstance(result, Struct)
 
 
 def test_to_lgdo_fields_present():
@@ -256,10 +234,10 @@ def test_to_lgdo_fields_present():
 def test_to_lgdo_array_types():
     sp = _make_superpulse()
     result = sp.to_lgdo()
-    assert isinstance(result["charge_wf"], LGDOArray)
-    assert isinstance(result["current_wf"], LGDOArray)
-    assert isinstance(result["charge_time_axis"], LGDOArray)
-    assert isinstance(result["current_time_axis"], LGDOArray)
+    assert isinstance(result["charge_wf"], Array)
+    assert isinstance(result["current_wf"], Array)
+    assert isinstance(result["charge_time_axis"], Array)
+    assert isinstance(result["current_time_axis"], Array)
 
 
 def test_to_lgdo_scalar_types():
@@ -275,7 +253,7 @@ def test_to_lgdo_scalar_types():
         "n_events_preliminary",
         "n_events_final",
     ):
-        assert isinstance(result[key], LGDOScalar), f"{key} should be LGDOScalar"
+        assert isinstance(result[key], Scalar), f"{key} should be Scalar"
 
 
 def test_to_lgdo_scalar_values():
@@ -311,126 +289,28 @@ def test_to_lgdo_waveform_values():
 
 
 # ===========================================================================
-# perform_data_selection
-# ===========================================================================
-
-
-def test_perform_data_selection_all_pass():
-    evt = _make_evt_data(n_events=10)
-    result = perform_data_selection(evt)
-    assert len(result) == 10
-
-
-def test_perform_data_selection_forced_trigger():
-    # Forced trigger events must be removed
-    evt = _make_evt_data(n_events=5, forced=True)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_pulser():
-    evt = _make_evt_data(n_events=5, puls=True)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_muon():
-    evt = _make_evt_data(n_events=5, muon=True)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_muon_offline():
-    evt = _make_evt_data(n_events=5, muon_offline=True)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_bad_channel():
-    evt = _make_evt_data(n_events=5, is_good_channel=False)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_not_bb_like():
-    evt = _make_evt_data(n_events=5, is_bb_like=False)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_multiplicity_gt_1():
-    evt = _make_evt_data(n_events=5, multiplicity=2)
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_low_spms_energy():
-    evt = _make_evt_data(n_events=5, spms_energy_sum=5.0)  # <= 10
-    result = perform_data_selection(evt)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_low_aoe_cut():
-    # low_aoe below threshold should be removed
-    evt = _make_evt_data(n_events=5, low_aoe=-5.0)
-    result = perform_data_selection(evt, aoe_low_threshold=-3.0)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_high_aoe_cut():
-    # high_aoe above threshold should be removed
-    evt = _make_evt_data(n_events=5, high_aoe=5.0)
-    result = perform_data_selection(evt, aoe_high_threshold=3.0)
-    assert len(result) == 0
-
-
-def test_perform_data_selection_mixed():
-    # 3 clean events + 2 forced triggers
-    clean = _make_evt_data(n_events=3)
-    dirty = _make_evt_data(n_events=2, forced=True)
-    evt = ak.concatenate([clean, dirty])
-    result = perform_data_selection(evt)
-    assert len(result) == 3
-
-
-def test_perform_data_selection_custom_aoe_thresholds():
-    # With a tighter threshold, even the "clean" default value (0.0) should pass
-    evt = _make_evt_data(n_events=5, low_aoe=0.0, high_aoe=0.0)
-    result = perform_data_selection(evt, aoe_low_threshold=-1.0, aoe_high_threshold=1.0)
-    assert len(result) == 5
-
-
-# ===========================================================================
 # select_detector_events
 # ===========================================================================
 
-TAB_MAP = {"V03422A": 1084803, "B00035B": 1104005}
-
 
 def test_select_detector_events_basic():
-    evt = _make_evt_data(n_events=10, rawids=[[1084803]] * 10)
-    result = select_detector_events(evt, "V03422A", TAB_MAP)
+    evt = _make_evt_data(n_events=10, detector_names=[["V03422A"]] * 10)
+    result = select_detector_events(evt, "V03422A")
     assert len(result) == 10
 
 
 def test_select_detector_events_filters_other_detector():
     # Mix 6 events for V03422A and 4 for B00035B
-    rawids = [[1084803]] * 6 + [[1104005]] * 4
-    evt = _make_evt_data(n_events=10, rawids=rawids)
-    result = select_detector_events(evt, "V03422A", TAB_MAP)
+    detector_names = [["V03422A"]] * 6 + [["B00035B"]] * 4
+    evt = _make_evt_data(n_events=10, detector_names=detector_names)
+    result = select_detector_events(evt, "V03422A")
     assert len(result) == 6
 
 
 def test_select_detector_events_none_match():
-    evt = _make_evt_data(n_events=5, rawids=[[1104005]] * 5)
-    result = select_detector_events(evt, "V03422A", TAB_MAP)
+    evt = _make_evt_data(n_events=5, detector_names=[["B00035B"]] * 5)
+    result = select_detector_events(evt, "V03422A")
     assert len(result) == 0
-
-
-def test_select_detector_events_unknown_detector():
-    evt = _make_evt_data(n_events=5)
-    with pytest.raises(KeyError):
-        select_detector_events(evt, "UNKNOWN", TAB_MAP)
 
 
 # ===========================================================================
@@ -824,7 +704,7 @@ def test_write_superpulses_lh5_structure(tmp_path):
 
     # The expected group path is V03422A/dt_900_1100_ns
     result = lh5.read("V03422A/dt_900_1100_ns", output_path)
-    assert isinstance(result, LGDOStruct)
+    assert isinstance(result, Struct)
 
 
 def test_write_superpulses_lh5_field_values(tmp_path):
@@ -862,5 +742,5 @@ def test_write_superpulses_multiple_slices(tmp_path):
     # Both groups must be present
     result1 = lh5.read("V03422A/dt_900_1100_ns", output_path)
     result2 = lh5.read("V03422A/dt_1100_1300_ns", output_path)
-    assert isinstance(result1, LGDOStruct)
-    assert isinstance(result2, LGDOStruct)
+    assert isinstance(result1, Struct)
+    assert isinstance(result2, Struct)
--- a/tests/test_superpulses.py
+++ b/tests/test_superpulses.py
@@ -2,22 +2,23 @@ from __future__ import annotations
 
 from pathlib import Path
 
+import awkward as ak
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 from lgdo import Array, Scalar, Struct, lh5
 
-from legendsimflow.scripts.build_superpulses_from_data import (
-    plot_chi2_cut,
-    plot_wfs_and_superpulse,
-)
 from legendsimflow.superpulses import (
     Slice,
     Superpulse,
+    _get_nested_field,
     compute_chi2,
     get_wfs_for_slice,
     lookup_wfs_indices,
-    write_superpulses_to_lh5,
+    plot_chi2_cut,
+    plot_wfs_and_superpulse,
+    read_superpulses,
+    write_superpulses,
 )
 
 dsp = {
@@ -132,6 +133,13 @@ def test_slice_is_frozen():
         sl.energy_range = (0.0, 1.0)
 
 
+def test_read_nested_field():
+    data = ak.Array({"geds": {"energy": [1, 2, 3]}, "spms_pe": [10, 2, 3]})
+
+    assert ak.all(_get_nested_field(data, "geds/energy") == ak.Array([1, 2, 3]))
+    assert ak.all(_get_nested_field(data, "spms_pe") == ak.Array([10, 2, 3]))
+
+
 def test_lookup_wfs_indices(legend_testdata):
     ref_path = legend_testdata.get_path("lh5/")
     path = ref_path / Path("l200-p16-r008-ssc-20251006T205904Z-tier_evt.lh5")
@@ -146,7 +154,12 @@ def test_lookup_wfs_indices(legend_testdata):
     ]
 
     file_idx = lookup_wfs_indices(
-        slices, evt_files=files, n_target=10, detector="V07302A"
+        slices,
+        evt_files=files,
+        n_target=10,
+        detector="V07302A",
+        t0_field="spms/first_t0",
+        end_time_field="geds/psd/drift_time",
     )
 
     assert isinstance(file_idx, list)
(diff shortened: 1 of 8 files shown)


workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+log = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Data structures
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class Pixel:
```
tdixon97: What is done by this class that cannot be achieved with simple datatypes?

I think this is a really nice guide https://learn.scientific-python.org/development/principles/design/

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+
+
+def _read_dsp_field_for_channel(
+    channels: ak.Array,
+    rows: ak.Array,
+    dsp_file: str,
+    field: str,
+    tab_map: dict[str, int],
+) -> ak.Array:
+    """
```
tdixon97: In general private functions prefaced by "_" have minimal docstrings, with the information instead added to the public (user facing) functions

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+
+    Raises
+    ------
+    ValueError
+        If any rawid in ``channels`` is absent from ``tab_map``.
+    """
+    raise NotImplementedError
+
+
+def select_data_in_pixel(
```
tdixon97: Is pixel really the best name?

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+
+    Raises
+    ------
+    KeyError
+        If ``detector`` is not found in ``tab_map``.
+    """
+    raise NotImplementedError
+
+
+def add_dsp_pars_to_evt(
```
tdixon97: I think that this should be replaced by just asking George to add these fields we need to the evt file, since in the end this code will be quite slow, take effort to maintain etc, and having one more field in evt files is no big deal.
Maybe you can make a note of the fields you need?

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+    -----
+    The ``geds.energy`` field (calibrated, in keV) is used for pixel selection
+    upstream. Normalisation here uses ``cuspEmax`` (uncalibrated, in ADC) so
+    that the charge waveform amplitude equals 1 at the plateau, independent of
+    gain calibration.
+    """
+    raise NotImplementedError
+
+
+def _subtract_baseline_and_normalise(
```
tdixon97: I think this and the following functions match quite closely existing code in dspeed.

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+
+    Returns
+    -------
+    ak.Array
+        Events within the pixel, shape ``(n_pixel_events,)``.
+    """
+    raise NotImplementedError
+
+
+def read_wfs_for_pixel(
```
tdixon97: I think if you do something like this
https://github.com/legend-exp/legend-simflow/blob/ee76d6ae344e7331b86f8f2e4000b5ba906c8bf7/workflow/src/legendsimflow/hpge_pars.py#L382
You can also avoid the waveform processing and dramatically reduce the amount of code.

less code is usually better

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+    """
+    raise NotImplementedError
+
+
+# ---------------------------------------------------------------------------
+# Orchestrator
+# ---------------------------------------------------------------------------
+
+
+def build_superpulse_for_pixel(
```
tdixon97: There are alot of kwargs here, luckily I think some of them are just parameters of the data processing and do not really need to be passed to the function, but they can be obtained from the data-flow config

If you have to keep them, I would make them keyword only

tdixon97:
This already looks good. I think one step thats helpful when designing the API is to write out a small script using it to see how the functions will interact. Again this example https://learn.scientific-python.org/development/principles/design/ is quite nice.

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
+"""
+This module implements the dataset preparation and superpulse construction
+for SSC data.
+
```
tdixon97:
```suggestion
"""
This module implements the dataset preparation and average ("superpulse") construction to characterize the pulse shape response of HPGe detectors.

This is an important step in tuning the pulse shape discrimination (PSD) simulations

```

maybe?

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1183 @@
...
+    Returns
+    -------
+    ak.Array
+        Concatenated EVT data with shape ``(n_events,)``.
+    """
+    raise NotImplementedError
+
+
+def perform_data_selection(
```
tdixon97: Could potientally be merged with the function before (while I generally support keeping IO/code seperate) the selection should be quite simple

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: So the dsp parameters I currently use are: 

- **tp_aoe_max** to calculate the drift time and build drift time slices. I think this could actually be added to evt.
- **cuspEmax** for the charge waveform normalization. I don't think this can really be added to evt, being just an uncalibrated parameter. I was using cuspEmax to minimize the impact of the noise on the normalization, but it is true that this dps parameters integration is the slowest part of all the workflow, so we could just normalize the waveform with min_max_norm after baseline subtraction. This would for sure make things simpler, but I am a bit worried about noise

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: Yes ok, I think _subtract_baseline_and_normalise,  _apply_mwa, _upsample_waveform and _compute_current can actually be removed as private functions and the corresponding operations can be done directly inside preprocess_waveforms.

I would keep _recalculate_tp_aoe_max and _align_and_trim because I don't see direct equivalents

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: Maybe a NamedTuple? But in the end it is still a class, no?
It's true that a "Pixel" is just two couples of numbers, but I think it is good to keep the names (energy and drift time), rather than accessing them by index.

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: EDTslice? edt_slice? I would call it just "slice" (both the class, "Slice" instead of "Pixel" and all the times it is mentioned in the code) but I think "slice" would suggest a 1-dimensional slice (a drift time slice in this context) and I didn't want us to forget we are doing also an energy slicing (which we should at a certain point question)

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: I see, it's true that this would basically replace the waveform preprocessing and do it in one line... 
My concern is that the WaveformBrowser processes one waveform at a time, and it returns either the charge or the current waveform, so we need to call it two times per waveform, and for all the 100/200 waveforms in each edt_slice. Also I think the alignment is done at 16ns sampling rate, while what I did was to recalculate t_Amax at 1ns and align by that.

I think if I replace all the private functions with just the corresponding dspeed functions inline in preprocess_waveforms, the code would already be much lighter than now, and I suspect that in the end it could be faster than calling the waveformBrowser. But I never really tried with it, so I cannot say for sure

tdixon97: > My concern is that the WaveformBrowser processes one waveform at a time, and it returns either the charge or the current waveform, so we need to call it two times per waveform, and for all the 100/200 waveforms in each edt_slice. Also I think the alignment is done at 16ns sampling rate, while what I did was to recalculate t_Amax at 1ns and align by that.

None of this is actually true:
* WaveformBrowser can process many WFs at once
* It can return multiple outputs
* It can also align at 1ns after MWA

It is really important we do not fill this repo with data processing code. I also really doubt its faster doing it yourself, since the underlying calculations are the same.
You can also alternatively of course just call `build_dsp` (basically equivlanet code)

SalehGiovanna: Ok I try with the waveformBrowser then

workflow/src/legendsimflow/superpulses.py
tdixon97: second could easily be extracted from the waveform via the WFBrowser

SalehGiovanna: Yess exactly, cuspEmax solved!
For now I would keep the add_dsp_pars_to_evt function just for the tp_aoe_max, otherwise the full module is unusable, but if/when it is added to evt I remove it

tests/test_superpulses.py
```diff
@@ -0,0 +1,853 @@
+"""Tests for legendsimflow.superpulses
+
+Covered functions
```
tdixon97: You can remove this, its unnecessary

tests/test_superpulses.py
```diff
@@ -0,0 +1,853 @@
...
+    select_detector_events,
+    write_superpulses_to_lh5,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers / shared fixtures
+# ---------------------------------------------------------------------------
+
+def _make_superpulse(
```
tdixon97: Make this a fixture

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+
+    args = parser.parse_args()
+
+    logger.info(
+        "--- Starting Superpulse Generation for %s (%s) ---", args.detector, args.runid
+    )
+
+    # Extract period and run — example: "l200-p16-r008-ssc" -> period="p16", run="r008"
+    try:
```
tdixon97: We have a function to do this, its in ` legendsimflow.metadata.parse_runid(runid)`

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,135 @@
...
+        period = parts[1]
+        run = parts[2]
+        run_type = parts[3]  # 'ssc', 'phy', or 'cal'
+    except IndexError as err:
+        msg = (
+            f"runid '{args.runid}' does not match expected format l200-period-run-type"
+        )
+        raise ValueError(msg) from err
+
+    # Data Files
```
tdixon97: This is is quite hardcoded, and it should be part of a function eg. see this https://github.com/legend-exp/legend-simflow/blob/611dd6f5ce68009283512663b80957dc9f02ea49/workflow/src/legendsimflow/scripts/extract_hpge_current_pulse_model.py#L160

This function should parse the data-prod config and extract the neccesary info. I think we already have what is needed

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Setup Metadata and Channel Map
+    lmeta = LegendMetadata(str(args.meta))
+    start_key = lmeta.datasets.runinfo[period][run][run_type]["start_key"]
+    chmap = lmeta.channelmap(on=start_key)
+    tab_map = {args.detector: chmap[args.detector]["daq"]["rawid"]}
+
+    # Retrieve the DSP config
+    logger.info("Retrieving DSP configuration...")
+    _, dsp_config, _, _ = lookup_currmod_fit_inputs(
```
tdixon97: it doesnt make sense why you only get the dsp config from this function, I think you want a function getting all the needed inputs.

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,135 @@
...
+            tab_map=tab_map,
+            evt_files=[str(f) for f in evt_files],
+            dsp_files=[str(f) for f in dsp_files],
+            raw_files=[str(f) for f in raw_files],
+            dsp_config=dsp_config,
+            dsp_fields=dsp_fields,
+            n_target_wfs=100,
+            chi2_threshold=3.0,
+        )
+
```
tdixon97, line 125: here we need some validation plots

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
+"""
+Module to implement the dataset preparation and average ("superpulse") construction to characterize the pulse shape response of HPGe detectors.
```
tdixon97, line 2: We do not need such a long module docstring, important things should be added to the docs

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
...
+
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+
+import awkward as ak
+import numpy as np
+from lgdo import Array as LGDOArray
+from lgdo import Scalar as LGDOScalar
+from lgdo import Struct as LGDOStruct
```
tdixon97: why ? I think it would be better not aliasing

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
...
+    ak.Array
+        Concatenated EVT data with shape ``(n_events,)``.
+    """
+    evt_arrays = []
+    for evt_file in evt_files:
+        evt_arrays.append(lh5.read_as("evt/", evt_file, library="ak"))
+
+    if len(evt_arrays) == 1:
+        return evt_arrays[0]
+    return ak.concatenate(evt_arrays)
```
tdixon97:
```suggestion
    return lh5.read_as("evt/", evt_files, library="ak")
```
function is unnecesary. In general this is very memory intensive, just read the fields you need

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
...
+        ak.all(evt_data.geds.quality.is_good_channel, axis=-1)
+        & ~evt_data.trigger.is_forced
+        & ~evt_data.coincident.puls
+        & ~evt_data.coincident.muon
+        & ~evt_data.coincident.muon_offline
+        & evt_data.geds.quality.is_bb_like
+        & (evt_data.geds.multiplicity == 1)
+        & (
+            evt_data.spms.energy_sum > 10
+        )  ### Is this actually needed? If not I could drop the spms entirely when reading evt files!
```
tdixon97: its needed but you should just only read `evt.spms.energy_sum` not the full spms ie

```python
lh5.read("evt",files,field_mask = ["geds","coincident","trigger","spms/energy_sum"])
```
will result in much less memory usage and much less time

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
...
+    psd_mask = ak.all(
+        evt_data.geds.psd.low_aoe.value > aoe_low_threshold, axis=-1
+    ) & ak.all(evt_data.geds.psd.high_aoe.value < aoe_high_threshold, axis=-1)
+    evt_data = evt_data[psd_mask]
+
+    log.debug("PSD cuts: %d -> %d events", n_after_preliminary, len(evt_data))
+    return evt_data
+
+
+def select_detector_events(
```
tdixon97: I think this can be merged with the line before, also note you have
```python
evt.geds.detector_name
```
in the evt files so you can do the selection without any rawid

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1385 @@
...
+    log.debug("%s: %d events", slice, len(slice_data))
+    return slice_data
+
+
+# ---------------------------------------------------------------------------
+# Waveform extraction (one slice)
+# ---------------------------------------------------------------------------
+
+
+def get_charge_and_current_wfs_for_slice(  ### Check entire function
```
tdixon97: This function is very verbose, and it returns alot of outputs, is that really needed?

tdixon97 (dismissed):

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+        for dt_start in range(500, 2500, 200)
+    ]
+
+    # Build superpulse for each slice
+    superpulses = {}
+    for current_slice in slices:
+        logger.info("\nProcessing %s...", current_slice)
+
+        sp = build_superpulse_for_slice(
```
tdixon97: I think its better for readability to have a few shorter functions but its just my preference. Eg. I would prefer something more like:

```
wfs = get_waveforms(...)

wfs_sel = similarity_select(wfs)

superpulse = build_superpulse(wfs_sel)
```
because each function is then a bit simpler

tdixon97:
Thanks Giovanna! I think its nice, but I also think the code is very complicated, can it be simplified? In the end the operations that are performed are not so complicated.
I guess some of the code was written with AI? In general the AI has a tendency to overdo things, and you sometimes have to tell it to simplify

tests/test_superpulses.py
SalehGiovanna: So I keep fix parameters only and I remove the tests in which I was giving parameters different from the default value (test_write_superpulses_multiple_slices)?

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: just to make it clear it was from lgdo when using, but ok removed

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: Are there any test evt files I can use to test it though? 
Without that i cannot test the selction either if it's just one function

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: Ok i switched to detector_name.

Not sure about moving the detector selection in the previous data selection function, it could be convenient to have all the good data first, and then to loop over detectors

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
SalehGiovanna: Ok i added a function (lookup_superpulse_inputs) here to reproduce that functions logic and return all the files I need.
Let me know if you had something like this in mind. If yes I will remove it from the script and place it somewhere else

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
SalehGiovanna: Yes I just tried with that to use the existing function, now I wrote one for this specific script.

workflow/src/legendsimflow/superpulses.py
SalehGiovanna: Update: i called it just "Slice"

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
SalehGiovanna: I would plot:

- Chi2 distribution
- Wfs in slice with superpulse
- Charge and current superpulses in different slices
- Other?

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1300 @@
...
+    raw_file: Path | str,
+    lh5_group: str,
+    indices: list[int],
+    dsp_config: Path | str,
+    charge_output: str = "wf_blsub",  ### What parameter should I use?? This? Or wf_av? Others?
```
tdixon97: i am not sure this the correct parameter to use, this is based on the presummed waveform. I think you want instead the windowed one

tdixon97:
@gipert this needs testing on the (new) ssc data, but is otherwise ready

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1076 @@
...
+
+
+@dataclass(frozen=True)
+class Slice:
+    """
+    Defines a 2D slice in the energy-drift-time space.
+
+    Parameters
+    ----------
+    energy_range : tuple[float, float]
```
gipert: i wonder if we need type hints also here or the docs render the ones below

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1076 @@
...
+class Superpulse:
+    """
+    Holds the average charge and current waveforms for one slice.
+
+    One ``Superpulse`` instance is produced per slice per detector, after the
+    full preprocessing and chi2 self-similarity cut. It carries both waveforms
+    together with the metadata needed to identify the slice, write to LH5, and
+    perform the subsequent electronics parameter optimisation.
+
+    Parameters
```
gipert: i would put the parameters section in the __init__() and remove the types

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1076 @@
...
+
+            - ``charge_wf``            : ``Array``, shape ``(n_charge_samples,)``
+            - ``current_wf``           : ``Array``, shape ``(n_current_samples,)``
+            - ``charge_time_axis``     : ``Array``, shape ``(n_charge_samples,)``, attrs ``{"units": "ns"}``
+            - ``current_time_axis``    : ``Array``, shape ``(n_current_samples,)``, attrs ``{"units": "ns"}``
+            - ``dt_center``            : ``Scalar``, drift time center [ns]
+            - ``dt_lo``                : ``Scalar``, drift time lower bound [ns]
+            - ``dt_hi``                : ``Scalar``, drift time upper bound [ns]
+            - ``e_lo``                 : ``Scalar``, energy lower bound [keV]
+            - ``e_hi``                 : ``Scalar``, energy upper bound [keV]
```
gipert: minor/clarity: e -> ene / energy

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1076 @@
...
+    mask = (
+        ak.all(evt_data.geds.quality.is_good_channel, axis=-1)
+        & (~evt_data.trigger.is_forced)
+        & (~evt_data.coincident.puls)
+        & (~evt_data.coincident.muon)
+        & (~evt_data.coincident.muon_offline)
+        & evt_data.geds.quality.is_bb_like
+        & (evt_data.geds.multiplicity == 1)
+        & (evt_data.spms.energy_sum > 10)
+        & (ak.all(evt_data.geds.detector_name == detector, axis=-1))
```
gipert, line 329: this is indeed slow, i wonder if using rawid would speed this up significantly

workflow/src/legendsimflow/superpulses.py
```diff
@@ -0,0 +1,1076 @@
...
+    tmp = data
+    for field_tmp in field.split("/"):
+        tmp = tmp[field_tmp]
+    return tmp
+
+
+def _select_data_in_slice(
+    det_evt_data: ak.Array,
+    drift_slice: Slice,
+    end_time_field: str = "geds/psd/drift_time",
```
gipert, line 350: an alternative to avoid the get nested field is to pass a lambda function: `lambda evt: evt.geds.psd.drift_time`

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
```diff
@@ -0,0 +1,261 @@
...
+    plot_wfs_and_superpulse,
+    write_superpulses,
+)
+
+MIN_NUMBER_WFS = 50
+TARGET_WFS = 200
+CHI2_THRESHOLD = 3.0
+
+CHARGE_OUTPUT = "wf_pz_bl_sub"
+CURR_OUTPUT = "curr_av"
```
gipert, line 50: this i guess is for a future pr, but the settings should go to legend-simflow-config

gipert (approved):
looks good to me, left some minor comments

workflow/src/legendsimflow/superpulses.py
tdixon97: i actually dont know

workflow/src/legendsimflow/superpulses.py
tdixon97: i think its fast enough for us

workflow/src/legendsimflow/superpulses.py
tdixon97: yeah maybe for the future, but for now what i did is working ok

workflow/src/legendsimflow/scripts/build_superpulses_from_data.py
tdixon97: yep lets add this in a PR on the integration

tdixon97:
I have added the script test and I fixed the associated issues it revelead. I think this is good to go now. I will make a second PR for the workflow integration

gipert:
can you clean up the git history? then i merge if ci is ok

tdixon97:
im just fixing some small issues then yes i can

tdixon97:
this is now ready pending the fixes in https://github.com/legend-exp/dspeed/pull/198

PR #199, opened 2026-04-09
fix: robustness fixes and full workflow integration test

gipert:
## Summary

- Fix missing SiPM handling in opt tier usabilities lookup
- Fall back to 1mm FCCD when diode metadata is missing instead of crashing
- Handle cylindrically asymmetric detectors in drift time map plots
- Skip NaN fraction check when there are too few entries (low statistics)
- Add `test_full_workflow` integration test exercising the full vtx→cvt pipeline at NERSC with real l200data
- Add console logging handler to test config for visible job output

## Test plan

- [x] `pre-commit run --all-files` passes
- [x] `pytest tests/test_utils.py::test_check_nans_leq` passes
- [x] `test_full_workflow` completes 19/19 steps successfully at NERSC


gipert:
right now this can only be run on NERSC, will gradually include in CI once we implement the possibility to run without optical map and l200data

PR #214, opened 2026-04-23
feat: add skip_opt and skip_hit settings to evt tier

### LINKED ISSUE #123: Do not run some parts

gipert:
## Summary

- Adds two optional boolean settings to the `evt` tier configuration: `skip_opt` (default `false`) and `skip_hit` (default `false`)
- When `skip_opt: true`, the `opt` (SiPM/LAr) Snakemake rule is skipped and the `evt` output contains only HPGe data (`geds`, `coincident/geds`); `spms` and `coincident/spms` are absent
- When `skip_hit: true`, the `hit` (HPGe) Snakemake rule is skipped and the `evt` output contains only SiPM data (`spms`, `coincident/spms`); `geds` and `coincident/geds` are absent; trigger fields are read from the `opt` tier instead
- Setting both to `true` simultaneously is a `WorkflowError`
- `pdf.py` detects available data from the cvt file structure at runtime and omits LAr-veto histograms (`mul_lar`, `mul_lar_psd`, `fail/lar`) when `coincident/spms` is absent
- Both settings are optional in the YAML (fall back to `false` when absent, so no existing configurations need updating)

## Test plan

- [ ] `pytest tests/tests/scripts/test_tier_evt.py::test_evt_script_cli_skip_both_raises` — no remage needed
- [ ] `pytest tests/tests/scripts/test_tier_pdf.py::test_pdf_script_cli_skip_opt` — no remage needed
- [ ] `pytest tests/tests/scripts/test_tier_pdf.py::test_pdf_script_cli_skip_hit` — no remage needed
- [ ] `pytest -m needs_remage tests/tests/scripts/test_tier_evt.py::test_evt_script_cli_skip_opt` — requires remage
- [ ] `pytest -m needs_remage tests/tests/scripts/test_tier_evt.py::test_evt_script_cli_skip_hit` — requires remage
- [ ] `pixi run test-python` — full suite, no regressions

### DIFF
--- a/docs/source/manual/meta.md
+++ b/docs/source/manual/meta.md
@@ -448,6 +448,8 @@ add_random_coincidences: false
 geds_energy_thr_kev: 25
 spms_energy_thr_pe: 0
 buffer_len: "50*MB"
+skip_opt: false
+skip_hit: false
 ```
 
 - `add_random_coincidences` (bool) — when `true`, random-coincidence (RC) SiPM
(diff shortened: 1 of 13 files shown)


tdixon97:
@gipert is this really what we wanted??

gipert:
yes?

tdixon97:
what i had in mind was skipping the psd simulations only, not the entire geds info

gipert:
sure we can add an additional hit config field to skip altogether, should be rather easy

but this PR still is very useful

PR #217, opened 2026-04-28
feat(pdf): per-detector-group histograms via configurable regex map

### LINKED ISSUE #203: Improve logging of LegendMetadata() version checkout

gipert:
## Summary

- Add an optional `detector_groups` block to the PDF tier settings that maps group names to Python regexes on LEGEND-200 detector names. The PDF tier emits one set of 1-D histograms per group at `pdf/<cut>/<group>/`. An implicit `all` group is always present.
- Cuts (good-channel, m=1, m=2, PSD single-site, LAr veto, fail masks) are computed globally and unchanged; per-group filtering only restricts which detector energies enter each group's histogram.
- The 2-D `mul2` histogram intentionally stays at the top level — per-group/per-pair 2-D splitting is out of scope for this iteration.
- The evt tier writes a top-level `detector_uids` LH5 `Struct` mapping detector names (union of hit and opt tiers) to reboost UIDs; cvt asserts agreement across inputs and propagates it. The PDF tier reads it once at startup, so it never deals with raw UIDs itself.
- `build_tier_pdf` gains a `params: detector_groups=...` block so config changes invalidate downstream outputs.

### Configuration

```yaml
# inputs/simprod/config/tier/pdf/<exp>/settings.yaml
detector_groups:
  icpc: "V.*"
  bege: "B.*"
```

### `detector_uids` layout in evt/cvt files

```
detector_uids · struct{V01,B02,...}
├── V01  · scalar<int64>     # reboost UID
├── B02  · scalar<int64>
└── ...
```

### Breaking change

PDF tier output schema changes from `pdf/<cut>` to `pdf/<cut>/<group>` (with `pdf/fail/<cut>/<group>` for fails). Downstream consumers must read the `all` group to preserve prior behaviour. External consumers (legend-pdfs, fits) are out of scope here.

## Test plan

- [x] `pixi run test-python` — full Python suite green (218 passed; pre-existing `test_dag` sandbox failure unrelated).
- [x] `pre-commit run --all-files` — clean on all touched files.
- [x] `cd docs && make` — clean build, no warnings.
- [x] Three new tests: schema with `detector_groups`, schema without (only `all`), and `sum(per-group) == all` cross-check on a non-overlapping `V.*`/`B.*` partition.
- [x] `caplog` assertion that the empty-group warning is silent on the `skip_hit` path.

### DIFF
--- a/docs/source/manual/meta.md
+++ b/docs/source/manual/meta.md
@@ -500,10 +500,23 @@ probability density functions from the concatenated event files.
 :caption: simprod/config/tier/pdf/{experiment}/settings.yaml
 
 buffer_len: "500*MB"
+
+# optional: split 1-D PDFs by detector group
+detector_groups:
+  icpc: "V.*"
+  bege: "B.*"
 ```
 
 - `buffer_len` (str) — LH5 read chunk size (e.g. `"500*MB"`). Controls memory
   usage during processing; does not affect the output.
+- `detector_groups` (mapping, optional) — maps group names to Python regex
+  strings. Each regex is matched against LEGEND-200 detector names using
+  `re.fullmatch`, so `"V.*"` selects all detectors whose names start with `V`.
+  When this key is absent, only the implicit `all` group is emitted (equivalent
+  to `detector_groups: {all: ".*"}`). Specifying `detector_groups` extends the
+  output: every named group is produced in addition to `all`, which is always
+  emitted regardless of the config. See {ref}`pdf-tier` for the resulting output
+  schema.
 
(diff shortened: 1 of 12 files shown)


gipert:
need to think about how to handle the uid -> name mapping, not sure this is the cleanest solution

tdixon97:
dont we have a function to extract this name to uid mapping?

gipert:
>dont we have a function to extract this name to uid mapping?

yes though we would need to add hit file as dependency of pdf, which is a bit ugly. the AI solution circumvents this

PR #218, opened 2026-04-28
feat(utils): symlink overridden paths into default locations

gipert:
## Summary

- Add `legendsimflow.utils.link_external_paths(config, workflow_basedir, *, logger=None)` that, when the user has manually overridden a `paths.<key>` entry in `simflow-config.yaml` to point at another production, creates a symlink at the canonical default location pointing to the override. Snakemake rules continue to read `config.paths.<key>` directly; the symlinks just keep this prod cycle's `generated/` tree navigable in the standard layout.
- Wire the call into `onstart:` in `workflow/Snakefile` so it runs once per workflow invocation.

## Configuration

Override one or more paths in your `simflow-config.yaml` exactly as before:

```yaml
paths:
  tier:
    hit: /scratch/other-prod/generated/tier/hit
    opt: /scratch/other-prod/generated/tier/opt
  pars: /scratch/other-prod/generated/pars
```

After the workflow starts you'll find:

```text
generated/tier/hit -> ../../../scratch/other-prod/generated/tier/hit
generated/tier/opt -> ../../../scratch/other-prod/generated/tier/opt
generated/pars     -> ../../scratch/other-prod/generated/pars
```

## Behaviour

Supported leaves: every `tier.<name>`, `pars`, `macros`, `geom`, `dtmaps`.

- For each leaf, if `config.paths.<key>` resolves to the canonical default (computed from the simflow's `templates/default.yaml` with `$_` → cwd), no symlink is created and any pre-existing one is removed (revert-to-local).
- Otherwise a relative symlink is created (or refreshed) at the default location pointing to the override.
- Real directories at the default location are never touched.
- `geom` / `dtmaps` defaults fall back to `<pars>/geom` and `<pars>/hpge/dtmaps`, mirroring `apply_path_defaults`.

Warnings are logged when:
- a real directory at the default shadows the override (override silently ignored);
- the override target doesn't exist (a broken symlink will be created);
- the override target is a non-directory.

The Snakefile passes the snakemake `logger`, so warnings are routed through the standard workflow logging channel.

## Test plan

- [x] `pixi run test-python` — full Python suite passes; 8 new tests in `tests/test_utils.py` cover override creation, no-op when at default, real-dir protection, idempotency, wrong-symlink replacement, stale-symlink removal, geom/dtmaps fallback, and intermediate-dir creation.
- [x] `pre-commit run --all-files` — clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ dependencies = [
     "dspeed",
     "pygama >=2.3.6",
     "pylegendmeta >=1.4.1",
+    "pyyaml",
     "reboost >=0.10.3,<0.11",
     "revertex >=0.1.2",
     "ruamel.yaml",
@@ -122,6 +123,7 @@ matplotlib = "*"
 numpy = "*"
 pylegendmeta = ">=1.4.1,<1.5"
 psutil = "*"
+pyyaml = "*"
 "ruamel.yaml" = "*"
 
 # execution
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import shutil
 import tempfile
 from pathlib import Path
 
@@ -11,6 +12,8 @@ from legendmeta import LegendMetadata
 from lgdo import Array, Table, lh5
 
 from legendsimflow import hpge_pars, utils
+
+_REPO_TEMPLATES = Path(__file__).resolve().parents[1] / "templates"
 
 
 def test_get_parameter_dict():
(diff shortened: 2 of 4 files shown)


gipert:
Addressed two of three Copilot suggestions in the amended commit (`a9135a2`):

- **Empty/comment-only template** → now early-returns when `yaml.safe_load` returns a non-dict (so the function stays a safe no-op as documented).
- **Misleading "real directory" wording** → changed to `"%s is an existing non-symlink path"` so the message matches both the regular-file and the directory case.

For the `import yaml` / PyYAML comment: this matches existing usage in `cli.py` and elsewhere in the simflow, where PyYAML is pulled in transitively (notably via Snakemake itself). Adding an explicit `PyYAML` dep is a separate, repo-wide cleanup; keeping this PR consistent with the surrounding code.

8 link tests still pass; pre-commit clean.

PR #222, opened 2026-05-04
Migrate from lgdo.lh5 to standalone legend-lh5io package

gipert:
## Summary
This PR migrates the codebase from using the `lh5` module bundled with `legend-pydataobj` to the standalone `legend-lh5io` package. This change aligns with the upstream decision to separate LH5 I/O functionality into its own package in `legend-pydataobj` v2.

## Key Changes

- **Dependency updates**: 
  - Added `legend-lh5io >=0.2` as a new dependency
  - Updated `legend-pydataobj` constraint from `>=1.17.1,<2` to `>=2`
  - Updated both `pyproject.toml` and `pixi.lock` dependencies

- **Import refactoring**: 
  - Changed all imports from `from lgdo import lh5` to `import lh5`
  - Updated imports of `LH5Iterator` from `lgdo.lh5` to `lh5`
  - Updated imports of datatype utilities from `lgdo.lh5.datatype` to `lh5.io.datatype`

- **API usage updates**:
  - Changed `lgdo.lh5.utils.read_n_rows()` to `lh5.read_n_rows()`
  - Updated datatype field access from `lgdo.lh5.datatype.get_struct_fields()` to `lh5.io.datatype.get_struct_fields()`

- **Documentation updates**:
  - Updated Sphinx intersphinx configuration to point to stable versions of both `lgdo` and new `lh5` documentation
  - Removed version pinning comment that was specific to v1.x compatibility

## Files Modified
- `pyproject.toml` and `pixi.lock`: Dependency updates
- `docs/source/conf.py`: Intersphinx configuration
- 25+ source and test files: Import and API usage updates across the codebase

## Implementation Details
All changes maintain backward compatibility at the functional level - the LH5 I/O operations remain identical, only the import paths and module organization have changed.

https://claude.ai/code/session_01LVUhn5SbWRgtbLgfd2ut4f

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,8 @@ dependencies = [
     "h5py",
     "hist",
     "legend-dataflow-scripts >=0.3.0a5",
-    "legend-pydataobj >=1.17.1,<2",
+    "legend-lh5io >=0.2",
+    "legend-pydataobj >=2",
     "legend-pygeom-l200 >=0.8",
     "legend-pygeom-tools >=0.2",
     "legend-pygeom-hpges >=0.9",
@@ -115,7 +116,8 @@ legend-simflow = { path = ".", editable = true }
 awkward = "*"
 dbetto = ">=1.3.2,<1.4"
 # legend-dataflow-scripts = "..."  # not on conda-forge
-legend-pydataobj = ">=1.17.1,<2"
+legend-lh5io = ">=0.2,<0.3"
+legend-pydataobj = ">=2,<2.1"
 legend-pygeom-l200 = ">=0.8,<0.9"
 legend-pygeom-tools = ">=0.2,<0.3"
 dspeed = ">=2.1,<2.2"
@@ -217,6 +219,7 @@ filterwarnings = [
   "error",
   "ignore:overflow encountered in divide:RuntimeWarning",
   "ignore:overflow encountered in cast:RuntimeWarning",
+  "ignore:lgdo.lh5 has moved to its own package:DeprecationWarning",
 ]
 log_cli_level = "INFO"
 testpaths = [
--- a/workflow/src/legendsimflow/tcm.py
+++ b/workflow/src/legendsimflow/tcm.py
@@ -19,9 +19,10 @@ from collections.abc import Iterable
 from pathlib import Path
 
 import awkward as ak
+import lh5
 import numpy as np
 import pygama.evt
-from lgdo import Table, lh5
+from lgdo import Table
 
 
 def build_tcm(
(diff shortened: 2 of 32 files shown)


ManuelHu:
claudes comment:

> legend-lh5io does not yet have a ReadTheDocs site, so the lh5 intersphinx inventory cannot be fetched.

this is wrong: https://legend-lh5io.readthedocs.io/en/stable/objects.inv

gipert:
You're right, sorry — the lh5io docs do exist; my sandbox couldn't reach `readthedocs.io` so I incorrectly assumed the inventory was missing. The actual issue is that the inventory documents the class at its definition path `lh5.io.iterator.LH5Iterator`, not the re-export `lh5.LH5Iterator`, so the cross-reference fails to resolve under nitpicky mode.

Fixed in 4a4fdcb: reverted to a proper `:class:` cross-ref using the canonical path and dropped the unnecessary `nitpick_ignore_regex` entries.

---
_Generated by [Claude Code](https://claude.ai/code)_

workflow/src/legendsimflow/tcm.py
```diff
@@ -148,7 +148,7 @@ def merge_stp_n_opt_tcms_to_lh5(
 ) -> None:
     """Stream-merge STP and OPT TCMs and write unified TCM to disk in chunks.
 
-    Iterates over `stp_file:/tcm` using :class:`~lh5.LH5Iterator`. For each
+    Iterates over `stp_file:/tcm` using `lh5.LH5Iterator`. For each
```
ManuelHu: `lh5.LH5Iterator` is just a re-export of` lh5.io.iterator.LH5Iterator
`
I _think_ sphinx just cannot resolve re-exported symbols?

gipert:
we should remove the deprecation warning ignore and merge this once the other LEGEND packages have been updated

pyproject.toml
```diff
@@ -217,6 +219,7 @@ filterwarnings = [
   "error",
   "ignore:overflow encountered in divide:RuntimeWarning",
   "ignore:overflow encountered in cast:RuntimeWarning",
+  "ignore:lgdo.lh5 has moved to its own package:DeprecationWarning",
```
ManuelHu, line 222: should we really filter the deprecationwarnings also for the current package? there is a way to selectively filter warnings just for some packages, so we could just filter the warnings from dependencies and not the simflow code...

ManuelHu: Otherwise we paint over new issues introduced by people still using the old scheme in new code. Adding `:pygama|reboost` to the end of the line should work

gipert: i would wait until all dependencies are fixed, then we can upgrade and remove the entire ignore

PR #238, opened 2026-05-27
Add realistic PSL validation plots

SalehGiovanna:


workflow/src/legendsimflow/scripts/make_hpge_realistic_pulse_shape_lib.py
```diff
@@ -44,6 +48,7 @@
         "electronics_model_file": "input.electronics_model",
         "input_file": "input.ideal_psl",
         "output_file": "output[0]",
+        "plot_file": "output.plots",
```
SalehGiovanna: @tdixon97 i added this based on the build_superpulses_from_data.py script, maybe check if it is correct or if something else should be modified

PR #243, opened 2026-06-01
generate superpulses on `cal` data for `phy` and allow for building PSL per run

tdixon97:
- [x] possibility to make a superpulse per run
- [x] modify electronics model for this option

### DIFF
--- a/workflow/src/legendsimflow/superpulses.py
+++ b/workflow/src/legendsimflow/superpulses.py
@@ -263,6 +263,17 @@ def lookup_superpulse_inputs(
         l200data = Path(l200data)
 
     _, period_int, run_int, data_type = mutils.parse_runid(runid)
+
+    # for physics runs we build superpulses from the reference calibration run:
+    # its raw/evt data and channel map are what we actually read below
+    data_runid = runid
+    if data_type == "phy":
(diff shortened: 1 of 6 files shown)


workflow/src/legendsimflow/superpulses.py
```diff
@@ -313,7 +321,7 @@ def _read_and_sel_evts(
             "coincident",
             "trigger",
             "spms/energy_sum",
-            "spms/event_t0",
+            "spms/first_t0",
```
gipert: this will print a warning in cal data, i can add an if

workflow/src/legendsimflow/superpulses.py
```diff
@@ -325,9 +333,11 @@ def _read_and_sel_evts(
         & (~evt_data.coincident.muon_offline)
         & evt_data.geds.quality.is_bb_like
         & (evt_data.geds.multiplicity == 1)
-        & (evt_data.spms.energy_sum > 10)
         & (ak.all(evt_data.geds.detector_name == detector, axis=-1))
     )
+    if "spms" in evt_data.fields:
+        mask = mask & evt_data.spms.energy_sum > 10
```
gipert: this needs to be updated to the new evt tier (spms event_t0)

PR #249, opened 2026-06-13
misc fixes

gipert:
- [x] fix: drop stale cache argument from archive_plots input
- [x] bump reboost and snakemake loggers versions
- [x] fix LegendSimflow.jl compat
- [x] fix dtmap plots targets

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,15 +67,15 @@ dependencies = [
     "pygama >=2.3.6",
     "pylegendmeta >=1.4.4",
     "pyyaml",
-    "reboost >=0.12.1,<0.13",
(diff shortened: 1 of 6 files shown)


gipert:
@tdixon97 I took the versions from `/global/cfs/cdirs/m2676/users/tdixon/sims/simflow/test-ssc-psl/workflow/src/LegendSimflow.jl/Manifest.toml`, but they are not ok as you can see from the CI

gipert:
Ok I think I found out

tdixon97:
Have you checked this doesnt force dtmaps to be generated if simulate_psd = False

gipert:
yes look at the code

PR #253, opened 2026-06-16
PSL fixes

gipert:
## Summary

Two independent sets of fixes on the PSL-based HPGe pipeline.

### HPGe realistic PSL memory / IO (hit tier)

- Load only the needed (000-degree) waveform axis of the realistic PSL and store the templates as `float32`, reducing peak memory in `build_tier_hit`
- Free the previous detector's PSL objects before loading the next one, lowering peak memory across the detector loop
- Write realistic PSL waveforms and drift-time maps as `float32`, reducing on-disk size, and handle degenerate A/E distributions when building histograms

### Electronics-model (sigma, tau) fit robustness

- Fix the slice-count guard in `extract_hpge_elec_response_model`: the truncation check compared `max_num_superpulses` against `len(ideal_wfs)` (the 4-key result dict) instead of the number of slices, and in the wrong direction
- Make the fit converge robustly to the minimum of the flat, correlated sigma-tau valley:
  - the raw cost is a mean RMS ~1e-3, so MIGRAD's scale-dependent convergence test stopped it early (e.g. ~1.8e-3 instead of ~6e-4 for V11924A); the objective is now internally scaled and MIGRAD runs with strategy 2
  - from a bad seed MIGRAD could land in the wrong basin; a coarse grid pre-scan (default 8x8, configurable via `grid_size`; `None` disables) now seeds it in the correct basin, removing the seed dependence
- Narrow `sigma_limits` to `(0, 100)` (the limits double as the grid span), and log the expected `invalid` status on the flat floor rather than treating it as a failure
- Add tests: grid recovers truth from a bad seed, `grid_size=None` skips the scan, and the best RMS is seed-independent

## Test plan

- [x] `pytest tests/test_hpge_electronics_tuning.py tests/scripts/test_extract_hpge_electronics_model.py` (16 passed)
- [x] pre-commit clean, docs build
- [ ] `build_tier_hit` runs without regressions and stays within the memory budget on NERSC

### DIFF
--- a/docs/source/manual/meta.md
+++ b/docs/source/manual/meta.md
@@ -400,6 +400,8 @@ absent, accessing any field raises an error.
 
 dead_layer_fraction: 0.5
 buffer_len: "500*MB"
+simulate_psd: True
+simulate_psd_with_psl: False
 
 eresmod_default:
   expression: FWHMLinear
@@ -431,6 +433,10 @@ psdcuts_default:
 - `aoeresmod_default` — A/E resolution model applied to detectors without a
   per-detector entry. See {ref}`build-tier-hit-hpge` for when this fallback is
   triggered.
+- `simulate_psd` a flag to perform the PSD simulations based on the single
+  template (default is `True`).
+- `simulate_psd_with_psl` a flag to perform the simulations based on the
+  pulse-shape-library (PSL), default: `False`,
 - `psdcuts_default` — PSD cut values applied to detectors without a per-detector
   entry. See {ref}`build-tier-hit-hpge` for when this fallback is triggered.
 
--- a/docs/source/manual/output.md
+++ b/docs/source/manual/output.md
@@ -26,17 +26,27 @@ These fields are carried over from the
 
 ### Added fields
 
-| Field             | Type    | Units | Description                                                                                                                                                                                                                                                                         |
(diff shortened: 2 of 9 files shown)


workflow/src/legendsimflow/hpge_electronics_tuning.py
```diff
@@ -277,12 +283,31 @@ def fit_electronics_parameters(
...
+    grid_size: tuple[int, int] | None = (8, 8),
 ) -> dict:
     """Fit the electronics response parameters sigma and tau.
 
     Minimises the mean RMS between simulated and measured current
     superpulses across drift-time slices using Minuit (MIGRAD).
 
+    The sigma-tau cost surface is a smooth but flat, correlated valley
+    (both parameters broaden the current pulse, so they trade off and are
+    poorly constrained by the 16 ns data sampling). A single MIGRAD run
```
tdixon97: the sampling in data is actually 1 ns ? We work with upsampled data  + sims

workflow/src/legendsimflow/hpge_electronics_tuning.py
```diff
@@ -277,12 +283,31 @@ def fit_electronics_parameters(
...
+    can land in the wrong basin from a bad seed, and (because the raw cost
```
tdixon97: I find some of this details a bit too much for a docstring

tdixon97:
I think it looks excellent

tdixon97:
@gipert is it simple to add the PSL extension here? 
It very clearly is a residual issue
image

PR #258, opened 2026-06-17
add energy aoe corr

tdixon97:


workflow/src/legendsimflow/scripts/tier/evt.py
```diff
@@ -435,10 +435,28 @@ def _read_hits(tcm_ak, tier, field):
                     "geds/psd/has_aoe", VectorOfVectors(~np.isnan(aoe[hitsel]))
                 )
 
-                is_ss = _read_hits(tcm, "hit", "is_single_site")
+                is_ss = _read_hits(tcm, "hit", "psd/is_single_site")
                 out_table.add_field(
                     "geds/psd/is_single_site", VectorOfVectors(is_ss[hitsel])
                 )
+                try:
```
tdixon97: @gipert we need to replace this with some check on `lh5.ls` but i wasnt sure how to make it work

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -379,6 +390,17 @@ def main() -> None:
                     )
                     aoe_res = aoeresmod_default(energy_true)
 
+                # get the mean A/E
+                if det_name in aoemean_func:
+                    aoe_mean = aoemean_func[det_name](energy_true)
+                else:
```
gipert, line 397: do we care about printing a warning if no aoe mean function is available?

tdixon97: maybe debug, i think it might be the case fairly often

workflow/src/legendsimflow/scripts/tier/hit.py
```diff
@@ -470,48 +489,64 @@ def main() -> None:
...
+                psd_sub_table.add_field(
                     "aoe",
                     lgdo.Array(np.asarray(psd_fields.aoe_class, dtype=np.float32)),
                 )
-                out_table.add_field(
+                psd_sub_table.add_field(
                     "is_single_site", lgdo.Array(psd_fields.is_single_site)
                 )
 
+                out_table.add_field("psd", psd_sub_table)
```
gipert: don't we want to be more explicit and have a `psd` subtable that contains subtables for each method?

workflow/src/legendsimflow/scripts/tier/evt.py
gipert: i think it's fine to just check in one hit table that the psd subtable exists

gipert:
need to update the docs before merging

workflow/src/legendsimflow/scripts/tier/hit.py
gipert: i found very useful to ask claude to read the logs when debugging stuff. so maybe for the future it's also ok if the logs are verbose

workflow/src/legendsimflow/scripts/tier/evt.py
tdixon97: yes but i always get confused with lh5.ls

workflow/src/legendsimflow/scripts/tier/hit.py
tdixon97: i think its ok like this, for me thats too many layers of nesting

gipert: what is the problem with nesting

workflow/src/legendsimflow/scripts/tier/evt.py
tdixon97: I decided to instead get this from the tier settings, I think thats more robust

PR #274, opened 2026-06-29
Prevent mean aoe being nan

tdixon97:


gipert:
see also minor code changes here: https://github.com/legend-exp/legend-simflow/pull/273

PR #287, opened 2026-07-03
test: add l200cfg09 NERSC workflow test with realistic PSL from data

### LINKED ISSUE #294: Update manual l200 NERSC integration tests
I.e. the pixi task `test-l200-workflow`

gipert:
Adds `test_l200cfg09_workflow`, a `needs_nersc` integration test that runs the full vtx→pdf pipeline for a new `l200cfg09` dummyprod experiment. Unlike the existing `l1000dsg01` test (which builds only the *ideal* PSL because it has no `l200data`), this experiment uses real `l200data` at `ref/v3.3.0` over two p16 `ssc` runs, so it exercises the realistic-PSL-from-data path: the pulse-shape library is built from data in the par tier, consumed by the hit tier into a `psd_psl` sub-table, and read back by the evt tier into `geds/psd_psl`. The test reuses the existing `_assert_psd_psl_in_hit` / `_assert_psd_psl_in_evt` helpers.

## What's included
- `tests/test_workflow.py`: new `test_l200cfg09_workflow` (`needs_nersc`, `needs_remage`)
- `tests/dummyprod/simflow-config-l200cfg09.yaml`: config entrypoint
- `tests/dummyprod/inputs/simprod/config/`: the `l200cfg09` config tree (geom, pars, and stp/opt/hit/evt/cvt/pdf tier settings)
- `tests/dummyprod/inputs/datasets/statuses/`: p15 status reset entry + status file
- `pyproject.toml`: `test-l200cfg09-workflow` pixi task

## Status
Draft: needs validation on NERSC (`pixi run -e test test-l200cfg09-workflow`). The test is NERSC-only and does not run in CI. Local `pre-commit run --all-files` passes and the non-NERSC suite is unaffected.


gipert:
the impurities for the dummy detectors chosen here are bad and dont make the field calculation converge

PR #308, opened 2026-07-11
feat: add energy correction script

tdixon97:
Needs a bit of work.
- [x] whether we also fit l200 data, its quite nice if available
- [x] logic for snakemake integration
- [x] output fileds
- [x] simid logic
- [x] what to do about runids... could have a global or runwise option (a bit like superpulses)
Regarding the data fitting, its definitely not essential and can be removed, but I think if we can keep it its rather nice because it should be easy and it should provide a nice validation plot for the MC.


tdixon97:
I see two approaches to implement:
1. a temporary hit tier with just the simids needed for this correction and then the final hit tier
2. a seperate tier for the fields that dont need "recalibration" from the hit tier, this is maybe a more elegant solution but we have to be careful of the proliferation of tiers

@gipert I like option 2), seems also more future proof and in general its good to split heavy computations from calibration. For the name I am not sure, but what about `dsp` since we produce A/E_raw ?

tdixon97:
@copilot this all lacks testing, can you write a simple script test, this should:
- generate simple fake hit tier (MC)
- run the script without l200 data
- check it runs and produces the right outputs

we could then also think about adding the test with l200data in a further step

tdixon97:
@copilot whats _FakeCalAoE and why is it neeed

gipert:
>a seperate tier for the fields that dont need "recalibration" from the hit tier, this is maybe a more elegant solution but we have to be careful of the proliferation of tiers

i think this adds quite some complexity and i am not sure we need it, expecially not now. i really think in the future we should study if this e- sim idea works, this would be much cleaner in terms of workflow, much faster, and also does not need to be tied to data parameters in principle (??? have to think about it)

so i think for this week we just need to be pragmatic and quickly integrate what we have for the ssc/rdc production, without aiming to have an elegant solution which we can work out in autumn. i don't want to spend too much time now on writing code that we will very likely change.

>what to do about runids...

we don't need this now? 

>Regarding the data fitting, its definitely not essential and can be removed, but I think if we can keep it its rather nice because it should be easy and it should provide a nice validation plot for the MC.

but you see this is a design choice, right now we never explicity load data just to make a plot and i think i kinda like this in a way because we try to keep the scope of the simflow limited and avoid not so necessary complexity. there are several other MC/data comparison plots that are equally interesting to produce. i think we should have this kind of plots but then i would not tie their production to the simflow. we can leave your data plotting code in for now but I would put it elsewhere in the future.

i can finish the snakemake integration today if we agree on this

docs/source/manual/meta.md
```diff
@@ -649,56 +674,11 @@ used at runtime.
...
-Each entry must provide a `single_template` and a `psl` block (one per PSD
-simulation type), and each block must contain:
-
-- `expression`: a Python expression for the A/E mean as a function of the energy
-  `x` (in keV), using the parameters `a` and `b`
-- `pars`: mapping of the parameters `a` and `b` to their values
-
-A detector with neither an explicit entry nor a `default` falls back to a flat
-A/E mean of 1 (a warning is logged for `on` detectors).
+The HPGe A/E mean energy-dependence correction is no longer supplied as
```
tdixon97: The no longer is not needed, just say what is done now

tests/scripts/test_get_aoe_energy_correction.py
```diff
@@ -0,0 +1,103 @@
...
+dummyprod = Path(__file__).parent.parent / "dummyprod"
+
+# per-method A/E width used to synthesize the fake hit data
+_SIM_TYPE_SIGMA = {"single_temp": 0.01, "pulse_lib": 0.012}
+
+
+def _write_fake_mc_hit(
+    path: Path,
+    det: str = "V05261A",
+    n_events: int = 8000,
```
tdixon97: might want a bit more to ensure no flaky rare pygama failures?

tdixon97:
should we merge?

gipert:
i would wait until we check the current prod

gipert:
this is not mergeable now from my point of view, needs some work still

tdixon97:
what issues?

PR #309, opened 2026-07-13
feat: setting up musun input

MoritzNeuberger:
Changes required to use the musun-gs input generated by revertex. See changes to legend-simflow-config [[here](https://github.com/legend-exp/legend-simflow-config/pull/38)].

Since musun-gs writes the position and kinematics into the same group (`vtx/kin`), the input logic for the generators had to be adjusted to use the `/RMG/Generator/FromFile/IncludePosition true`.

I set up logic to determine whether the generator input file contains position information within the same group. With this logic, users do not have to worry or add additional flags for it to use such input.


MoritzNeuberger:
Additional PRs with further implementations needed to run the muon simulation will follow. This PR is just about getting the muon simulations to run.

gipert:
mmmh not sure i like this approach, why don't we change the metadata API to make this cleaner? we can distinguish between multiple vertices schema with multiple keywords, not only `~vertices:`

MoritzNeuberger:
I'm ok with `~vertices` since it directly indicates that a vtx file is generated. Maybe one could add a flag to the specific entry in `vtx/.../simconfig.yaml` indicating the kind of output it generates. In macro generation, one can access this information then via `mutils.get_vtx_simconfig`. This would be a manual approach, but since we have only a single generator with this kind of output so far, I'd be fine with it. Was this what you imagined?

gipert:
my general comment about the approach is: in general i prefer to ask the user to be more specific in some cases to avoid having to write weird/obscure code that infers things and can be fragile / hard to maintain.

i think what you say sounds good, but unfortunately i don't have time to look at this more carefully now to give you a better advice!

MoritzNeuberger:
I implemented the above suggestions. Now every `vtx/.../simconfig.yaml` contains a `product` key that specifies which type of content it produces (`position`, `kinematic`, `kinematic_and_position`). I am open to renaming it, it was just the first idea I had.

Based on this entry, `make_remage_macro` now has different branches in the generator and confinement sections depending on the required commands. In addition, it also throws an error when incompatible options are selected, also covered by the tests.

PR #313, opened 2026-07-21
Feat: implement logic for the maurina gamma cascades

MoritzNeuberger:
Add logic to automatically pull the [[generated-maurina-output](https://github.com/legend-exp/generated-maurina-output/tree/main)] repository and generate the macro commands to link into the remage simulation. Adds a rule that runs a `git clone` command with `--depth 1` since the repository is ~1Gb. This requires a new entry in `config.paths` to specify where the repository should be cloned (opinions?). Users can choose to run the simulation with the generated cascades by specifying `maurina_gamma_cascade` in the stp simconfig.yaml for the specific simid. This can be either a boolean indicating whether all or none (default) are registered, or a list of tuples specifying which isotopes should be registered.

### DIFF
--- a/docs/source/manual/meta.md
+++ b/docs/source/manual/meta.md
@@ -155,6 +155,10 @@ Supported fields per `simid`:
   for this `simid`. This configuration block is injected unmodified to the
   geometry tooling (currently
   [legend-pygeom-l200](https://legend-pygeom-l200.readthedocs.io)).
+- `maurina_gamma_cascades`: Optional boolean (default `false`). If `true`,
+  register the tabulated MAURINA gamma cascades with _remage_. Every isotope
+  directory in the cascade data repository is registered, see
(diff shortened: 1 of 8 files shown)


EricMEsch:
Looks good. Question: Do we also want to add a file for Hydrogen? As the Hydrogen capture is completly crazy in Geant4.11.3 (see the remage neutron validation) and Geant4.11.4 where they fixed it, the Hadronic physicslists are broken...

image

EricMEsch (approved):

MoritzNeuberger:
I have no issues with adding it. In the end, it would be a single maurinaesc file with a single entry, i.e., the 2.2MeV gamma, right? Please go ahead adding in the corresponding repo.

ManuelHu (changes requested):
you will still need to change the version bound of remage, right? (if we would already have a released version of remage with this...)

Issue #117, opened 2026-02-26
Ideas about generalization to accomodate L1000


MoritzNeuberger:
I'm curious what your thoughts are on integrating L1000 sims into simflow. 

To use the infrastructure of snakemake as is right now, we would need to set up some mock `metadata`/`l200data` repos to use as a stand-in. This could be useful to have a central place to define our expectations for the experiment, and it can be generalized in the future (e.g., have several mock detectors with different parameters, etc.).

I am wondering which parts of the post-processing are not sufficiently generalizable to L1000. The only thing that comes to mind is the optical information. Opt could be called several times for each relevant volume (ULAr, ALAr). Evt can only read a single opt file, but can you merge the opt files? Otherwise, the generalization of evt should be straightforward. I guess this would also be relevant if one wants maps for PEN plates?

gipert:
>we would need to set up some mock metadata/l200data

This is something that I would like to design carefully, it's also quite useful for the L200 simflow too and we should not jsut uncontrollably increase the amount of branching in the code

gipert:
WIP: I'm currently working on a series of PRs that will allow us to override all the pars with metadata. part of the work has already been merged and you can have a look at the updated docs

MoritzNeuberger:
I am reopening this issue since there are still open points that need to be addressed. Here is a summary of what I found:

1. **geometry generation**: at the moment, the geometry generation in [[link](https://github.com/legend-exp/legend-simflow/blob/a27b1a42d0bf3ad14bb30537b8bf0d93525d8f6c/workflow/rules/stp.smk#L86)] is hardcoded to the l200 geometry. We need to implement logic to determine which geometry to use, maybe by matching the `experiment` name to `l200` or `l1000`? In addition, the cli of l1000 currently lacks a `--config` option because I changed it some time ago to differentiate between high-level (->which part to implement, see [[link](https://legend-pygeom-l1000.readthedocs.io/en/latest/runtime-cfg.html#raw-configuration-files)]) and low-level (->channelmap+special metadata, see [[link](https://legend-pygeom-l1000.readthedocs.io/en/latest/runtime-cfg.html#compilation)]) config files that one could feed it (I called them *raw* and *compiled*, though I admit those names are awkward). 
2. **channel map / special metadata**: as mentioned above, the geometry generator also generates a dummy channel map and special metadata. These could be referenced directly w/o the need for additional files elsewhere.
[...]

PR #319, opened 2026-08-05
feat: implementing functionality to run l1000 simulations (see #117)

MoritzNeuberger:
This PR introduces some changes to make LEGEND-1000 simulations run all the way to pdf tier (see #117). These changes consist of making several pieces that were hard-coded more generalized to 200. Here is an overview:

- experiment_prefix derives the runid prefix from the experiment name
- get_hpge_voltage accepts a default entry in the opv database, so 336
  detectors need one line instead of 336
- the evt tier reads the scintillator volume name from the opt tier settings
  instead of the hardcoded liquid_argon
- the LAr veto thresholds move to the evt tier settings, with the LEGEND-200
  values as the default
- gen_geom_config writes absolute paths, because the geometry generators
  resolve relative paths against their own directory

The biggest change is the option to use generated metadata from tarballs provided with geom-config. Since L1000 does not have its own metadata, it has to be generated separately. Since their content is heavily dependent on the specific geometry implementation, this generator was implemented in the L1000 geometry generator (see [[pr](https://github.com/legend-exp/legend-pygeom-l1000/pull/132)]). This generates the mentioned tarball of the generated metadata that can be used (1) to generate the corresponding geometry (since it also contains the `special_metadata.yaml`), but (2) it can also be directly extracted into the `paths.metadata` folder as a direct replacement for the l200 metadata. Since the tarball is shipped together with the geom-config, the workflow looks as follows:
- If a `generated_metadata` entry is set to `true` in the `simflow-config.yaml`, `bootstrap_generated_metadata` is run before the LegendMetadata is initialized
- It first clones the simprod repo into `paths.metadata` and then looks for the specified experiments geom-config and thus the metadata tarball `paths.metadata`. To avoid writing the generated metadata every time, a checksum stamp is saved in `paths.metadata` that compares to the tarball's content, and if identical, the extraction is skipped
- Everything after this should work as before, since it imitates the metadata like the l200 one

I am currently preparing a PR for the simflow config for a L1000 setup. It heavily relies on the defaults previously introduced by Luigi based on the dummy prodenv in the tests.


MoritzNeuberger:
Refactor # 2: read experiment metadata from an overlay in simflow-config 

We want to treat legend-metadata as static. Therefore, we should not copy or adjust it in any way. In this refactor, the simflow does not generate anything into it. An experiment that legend-metadata does not describe brings its own metadata in legend-simflow-config instead. The metadata overlay lives under `input/simflow/config/metadata/<experiment>/` and uses the legend-metadata layout. The idea is that simflow first tries to find referenced information in the regular legend-metadata, and if it fails, it falls back to search for it in the provided overlay.

`metadata.lookup` drives the fallback. It applies the given query for a parameter to legend-metadata first and to the overlay second. Every accessor uses it: `get_runinfo`, `get_channelmap`, `get_diode`, `get_crystal`, `get_runlist` and `simpars`. The rest of the Python code therefore does not care from which database a parameter was taken.

Rules that need a file on disk cannot use `lookup`. `patterns` resolves those paths against the same two roots. The two Julia pulse-shape rules now get the database that holds their detector as `--metadata`.

The runs of the overlay must stay clear of the legend-metadata runs, because the simflow queries legend-metadata first:

  - Number the periods outside the range of the real experiment, for example p99. The simflow resolves a run by period and run number alone.
  - Keep the start keys before the legend-metadata era, for example the year 2000. The simflow resolves a channel map by timestamp, so the period number alone does not cover it. A timestamp that precedes every validity entry raises an error. The lookup then reaches the overlay.

`validate_metadata_overlay` refuses a colliding period at startup.

The dummy production runs on a committed overlay with p99 runs. Each l1000dsg01 parameter directory gained a matching validity entry.

Another idea was to provide flat yaml files as reference values instead of providing metadata overlay in the usual tree format of the metadata. This would require some more significant rewriting since then the query functions are also different for each accessor. In addition, the Julia scripts expect the data to be in this format and would require rewriting.

**Next**: I will split this pr into two. 1) everything up to stp tier (geometry generation, etc.) which is rather uncontroversial. 2) is this pr that takes care of the metadata overlay.

PR #323, opened 2026-08-18
fix: tolerate missing HPGe flags when caching detector usabilities

rosannadeckert:
**Issue:**
`cache_detector_usabilities` fails with `KeyError: 'psd_usability'` for any experiment without HPGe detectors.

`aggregate.pivot_detinfo()` pivots `{runid: {det: {flag: value}}}` into `{flag: {runid: {det: value}}}`, and — as its own docstring states — a flag absent from every entry simply does not appear in the result.
`gen_list_of_all_usabilities()` sets `psd_usability` and `crystal_metadata_usability` only for detectors with `system == "geds"`, so a LAr-only configuration produces a mapping containing `usability` alone. The rule then indexes all three keys unconditionally and raises.

I hit this issue while setting up a LAr-only production (p13, no germanium detectors installed), where it blocks the `opt` and `evt` tiers since both depend on the cached usabilities.

**Fix:**
Index the pivoted mapping with `.get(flag, {})` and write the HPGe-specific files empty when no detector carries those flags. Experiments with HPGe detectors are unaffected: the keys exist, so `.get()` returns exactly what the subscript returned before.

Note: To truly test this, we'd need some dummy data without Ge detectors.

### DIFF
--- a/workflow/rules/aux.smk
+++ b/workflow/rules/aux.smk
@@ -205,10 +205,14 @@ rule cache_detector_usabilities:
         detinfo = aggregate.pivot_detinfo(
             aggregate.gen_list_of_all_usabilities(config).to_dict()
         )
-        dbetto.utils.write_dict(detinfo["usability"], output.usability)
-        dbetto.utils.write_dict(detinfo["psd_usability"], output.psd_usability)
+        # `pivot_detinfo` only creates a key for a flag that at least one
+        # detector carries, and psd/crystal usability are set for HPGes only.
+        # An experiment without HPGe detectors therefore yields neither, so
+        # index defensively and write those files empty instead of raising.
+        dbetto.utils.write_dict(detinfo.get("usability", {}), output.usability)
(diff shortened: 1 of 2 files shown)


workflow/rules/aux.smk
```diff
@@ -205,10 +205,14 @@ rule cache_detector_usabilities:
         detinfo = aggregate.pivot_detinfo(
             aggregate.gen_list_of_all_usabilities(config).to_dict()
         )
-        dbetto.utils.write_dict(detinfo["usability"], output.usability)
-        dbetto.utils.write_dict(detinfo["psd_usability"], output.psd_usability)
+        # `pivot_detinfo` only creates a key for a flag that at least one
+        # detector carries, and psd/crystal usability are set for HPGes only.
+        # An experiment without HPGe detectors therefore yields neither, so
+        # index defensively and write those files empty instead of raising.
```
gipert:
```suggestion
```

PR #328, opened 2026-08-21
feat: build the geometry with the generator the experiment names + small adjustments that are geometry name dependent

MoritzNeuberger:
The geometry generator is a specified as a string in the experiment setups geom-config. LEGEND-200 uses `legend-pygeom-l200` and LEGEND-1000 uses `legend-pygeom-l1000`. The template geometry configuration file of the experiment now names its generator in an `executable` field. `geometry.geom_executable` reads that field and falls back to `legend-pygeom-l200`. `build_geom_gdml` runs the command it returns, and `render_geometry` imports the matching package.

A LEGEND-1000 geometry names its scintillator volume `undergroundlar`, and not `liquid_argon`. The opt tier settings therefore supply the name, and the evt tier reads it from there. The evt tier raises a clear error when the volume is absent from the stp file. The two LAr veto thresholds become evt tier settings as well. Their defaults are the LEGEND-200 values, so the behavior does not change.

The l1000dsg01 dummy production now builds a real LEGEND-1000 geometry. Its stp confinement matches the LEGEND-1000 insulator volumes.

### DIFF
--- a/tests/scripts/conftest.py
+++ b/tests/scripts/conftest.py
@@ -129,17 +129,48 @@ def legend_dtmap_path():
     return testprod / "inputs/simprod/V05261B-4200V-hpge-drift-time-map.lh5"
 
 
-def _l1000_config(tmp_dir: Path) -> Path:
+#: opt tier settings the fixtures below need on top of the committed ones.
+#: `legend_gdml_path` builds the geometry with legend-pygeom-l200, whose
+#: scintillator volume is `liquid_argon`. The committed settings name the
+#: LEGEND-1000 volume instead, which only `test_l1000_workflow` produces.
+_L1000_SETTINGS_OVERRIDE = {"opt": {"scintillator_volume_name": "liquid_argon"}}
+
+
+def _l1000_config(tmp_dir: Path, settings_by_tier: dict | None = None) -> Path:
     """Write a minimal simflow-config-l1000.yaml to *tmp_dir* and return its path.
 
-    Only ``paths.metadata`` is overridden to point at the dummyprod inputs;
-    all other path entries use ``$_`` substitution resolved to *tmp_dir*.
+    The dummyprod metadata is copied under *tmp_dir* so the tier settings can be
+    edited without a change to the committed tree. ``settings_by_tier`` maps a
+    tier name to the settings keys to overwrite, on top of
+    :data:`_L1000_SETTINGS_OVERRIDE`. All path entries other than
+    ``paths.metadata`` and ``paths.config`` use ``$_`` substitution resolved to
+    *tmp_dir*.
     """
+    meta = tmp_dir / "inputs"
+    shutil.copytree(testprod / "inputs", meta)
+
+    overrides = {tier: dict(o) for tier, o in _L1000_SETTINGS_OVERRIDE.items()}
+    for tier, overlay in (settings_by_tier or {}).items():
+        overrides.setdefault(tier, {}).update(overlay)
+
+    for tier, overlay in overrides.items():
+        f = meta / "simprod/config/tier" / tier / "l1000dsg01/settings.yaml"
+        data = yaml.safe_load(f.read_text())
+        data.update(overlay)
+        f.write_text(yaml.safe_dump(data))
+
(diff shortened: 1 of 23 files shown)


MoritzNeuberger:
This is a smaller pr to implement what was initially in #319

workflow/src/legendsimflow/geometry.py
```diff
@@ -31,6 +31,9 @@
 # pygeomtools, pygeoml200) are imported lazily inside the two functions below,
 # not at module level, so the module stays light for `load_vis_scene` and tests.
 
+# geometry generator invoked when the experiment's template geometry
+# configuration file does not name one explicitly; see `geom_executable`.
+DEFAULT_GEOM_EXECUTABLE = "legend-pygeom-l200"
```
gipert: i don't think this is needed, we should just be explicit

workflow/src/legendsimflow/geometry.py
```diff
@@ -56,6 +59,74 @@
     },
 }
 
+DEFAULT_VIS_SCENE_L1000: dict = {
```
gipert: maybe also these vis defaults are not needed. should we just always provide the full config in the metadata and reduce code here?

workflow/src/legendsimflow/geometry.py
```diff
@@ -56,6 +59,74 @@
...
+
+    for field in GEOM_CONFIG_PATH_FIELDS:
+        value = resolved.get(field)
+        if isinstance(value, str) and not Path(value).is_absolute():
+            resolved[field] = str(root / value)
+
+    return resolved
+
+
+def geom_executable(config: SimflowConfig) -> str:
```
gipert: i don't think this needs a function. also in general claude is too verbose with the docstrings and ends up writing useless functions like this

workflow/src/legendsimflow/geometry.py
```diff
@@ -56,6 +59,74 @@
...
+    },
+}
+
+
+#: geometry configuration fields that can hold a filesystem path.
+#: :func:`resolve_geom_config_paths` makes these fields absolute.
+GEOM_CONFIG_PATH_FIELDS = ("metadata", "raw_config", "channelmap", "special_metadata")
+
+
+def resolve_geom_config_paths(geom_config: Mapping, source: str | Path) -> dict:
```
gipert: i don't understand the need of this new function?

MoritzNeuberger: I am confused. I do not have it present in my local branch. I think it is part of an older implementation and should not be here. Lets see what happens when I push the changes in a min.

MoritzNeuberger:
The error in the tests occur due to missing infos in the metadata. This PR is only supposed to address the geometry generation such that simulations can run up to stp tier. Should I revert the l1000 tests such that they are based on l200 again to avoid the tests to fail?

tests/scripts/conftest.py
```diff
@@ -129,19 +129,50 @@ def legend_dtmap_path():
     return testprod / "inputs/simprod/V05261B-4200V-hpge-drift-time-map.lh5"
 
 
-def _l1000_config(tmp_dir: Path) -> Path:
+#: opt tier settings the fixtures below need on top of the committed ones.
+#: `legend_gdml_path` builds the geometry with legend-pygeom-l200, whose
+#: scintillator volume is `liquid_argon`. The committed settings name the
+#: LEGEND-1000 volume instead, which only `test_l1000_workflow` produces.
+_L1000_SETTINGS_OVERRIDE = {"opt": {"scintillator_volume_name": "liquid_argon"}}
```
gipert: claude likes to make these global variables but what is the need for this?

MoritzNeuberger: remove it and in general removed some slop

MoritzNeuberger:
Added a new requirement for the simflow-config: `lar_veto_multiplicity_thr` and `lar_veto_energy_sum_pe_thr` now have to be set explicitly in the evt config. This is to avoid setting defaults for these cuts threshold in the simflow source code.

Also prepared a [PR](https://github.com/legend-exp/legend-simflow-config/pull/42) in simflow-config specifically for this PR.

PR #330, opened 2026-08-24
feat: implementing metadata fallback for parameters not present in _legend-metadata_

MoritzNeuberger:
Experiments not represented by _legend-metadata_ require additional option for input. This PR implements a workaround using the new `metadata.lookup` function and an additional folder in `legend-simflow-config` called `metadata`. Here an example for the L1000 geometry.

```
├── geom
│   ├── l1000dsg01-geom-config.yaml
│   ├── l1000dsg01-vis-config.yaml
│   ├── ...
├── metadata
│   └── l1000dsg01
│       ├── datasets
│       │   ├── runinfo.yaml
│       │   ├── runlists.yaml
│       │   └── statuses
│       │       ├── l1000-p01-r%-T%-all-config.yaml
│       │       └── validity.yaml
│       └── hardware
│           ├── configuration
│           │   └── channelmaps
│           │       ├── l1000-p01-r%-T%-all-config.yaml
│           │       └── validity.yaml
│           └── detectors
│               └── germanium
│                   ├── crystals
│                   │   └── V011.yaml
│                   └── diodes
│                       ├── V00101X.yaml
│                       ├── V00102X.yaml
│                       ├── V00103X.yaml
│                       ├── V00104X.yaml
│                       ├── ...
├── pars
└── tier
```

The entries in the `metadata` folder have the same format as _legend-metadata_. Whenever simflow tries to look up a parameter from the metadata, it uses `metadata.lookup` as a wrapper around the query that searches for the parameter, e.g., `get_runinfo`, `get_channelmap`, etc. It does the lookup in a two step process: First it checks _legend-metadata_ for the parameter, and if not found, it checks the new metadata folder in `legend-simflow-config` of the specified experiment setup.

Since the additional metadata entries are only considered when the original _legend-metadata_ does not contain the queried parameter, I would suggest calling the concept **_fallback-metadata_**. I admit, while being technically correct, there might be better names describing what it is supposed to do? _supplementary-metadata_? _standin-metadata_? In any case I used this terminology through out the implementation and docs for now.

The motivation for the _fallback-metadata_ to have the same format as _legend-metadata_ is twofold:
1. we can reuse the same queries for parameters as for _legend-metadata_ requiring us to write less additional logic.
2. the julia part requires a `--metadata` path to access the required parameters separately. Thus any additional logic would also have to be reproduced in julia.

In simflow, the _fallback-metadata_ is only used for querying parameter. The geometry is entirely generated based on the content of `XXX-geom-config.yaml`. 

An example of how the _fallback-metadata_ looks like for L1000 can be seen in [this](https://github.com/legend-exp/legend-simflow-config/pull/41) PR in legend-simflow-config.

This PR superseeds #319 which I will close now.

**Sidenote**: the branch is currently based on the one from #328 , therefore also some geometry related implementations have leaked in. I though this to be easier for merging later, though now its difficult to separate by eye. If requested, I can rebase it again on main.


MoritzNeuberger:
Some more comments:
- The changes in most files consist of the fact that the parameter lookup now requires the config object rather than the metadata object
- The most relevant changes occurred in `metadata.py`. I admit that Claude generated a lot of guard rails that might not be entirely needed. If requested, I will clean this up a bit.
- The documentation also describes some requirements on the format of the _fallback-metadata_, i.e., the timestamps being before the first l200 run (e.g. 1.1.2000) or the period being p99. In addition, to avoid collisions of generated HPGe detectors, in legend-pygeom-l1000 I set the slice parameter for the generated HPGe to "X", i.e., as shown in the tree above, they all have the format `VXXXYYA` with XXX being the string and YY the position. See [PR](https://github.com/legend-exp/legend-pygeom-l1000/pull/132).
- Some tests failed on the CI that didn't on my machine. Will investigate.

MoritzNeuberger:
I woke up this morning and was horrified by the state that I thought I could create this pr with. I will rebase on main and clean up a bit.

MoritzNeuberger:
Rebased on main. 

Another note that might be of interest: All references to the metadata are now wrapped in the metadata.py. This means there should be no more direct references to the metadata anywhere else in the package. I didn't verify this but I think this is anyhow a good thing.

gipert:
We decided to create a legend1000-metadata repo instead. It's cleaner and we establish a convention for simulating experiments with the Simflow

